PR #16, opened 2023-04-25
feat: Create Indexer for transactions

### LINKED ISSUE #12: Create Indexer for transactions
[FRO-26](https://linear.app/fuel-network/issue/FRO-26/create-indexer-for-transactions)

Enable users to list all transactions with;

* Tx ID
* status
* Age
* Address involved
* From
* To

Braqzen:
## Changes

- There is 1 transaction query target which consists of 3 transaction types
  - `Script`
  - `Create`
  - `Mint`
- The transaction type in the schema is pruned meaning it does not contain all available fields / data from the client. Additional fields may be added in the future
- The transaction consists of two optional fields:
  - Success state (minting does not have one)
  - UTXO Inputs (minting does not have inputs)
- The transaction has 3 mandatory fields
  - ID used by the indexer
  - Age is the block time
  - Outputs
- The schema type do not currently use strict types and thus the following fields use `Json` for now
  - Success state
  - Inputs
  - Outputs

## Notes

- WIP: awaiting indexer features to allow the crate to compile and use strict types

## Issue

Closes #12

### DIFF
--- a/packages/indexer/schema/indexer.schema.graphql
+++ b/packages/indexer/schema/indexer.schema.graphql
@@ -3,79 +3,27 @@ schema {
 }
 
 type QueryRoot {
-  block: Block
   tx: Tx
-  account: Account
-  contract: Contract
 }
 
-# https://ethereum.org/en/developers/docs/data-and-analytics/block-explorers/
-
-type Block {
+type ScriptTx {
   id: ID!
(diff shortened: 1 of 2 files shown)


Braqzen:
Closing in favour of #19 because it requires transaction to be indexed as well.

PR #27, opened 2023-08-05
feat: add explorer graphql

luizstacio:
This PR creates a GraphQL API that returns the data from fuel-core and extends it with metadata. This approach will be a base for the feature work when we include more metadata like ABIs and code parsing.

This is a different approach from the Federation. The problem with the federation is that it incentives a chain of queries. 

Ex.: If we want to know the tokens and accounts from related data from a transaction list, the interface would be required to;
1. Execute a request to load the transactions
2. Create logic to get assetIds and Accounts it wants to have the metadata
3. Execute another request to load the accounts and tokens metadata

With the approach from `Schema Stitching`, we can chain this data directly on the server, offloading the logic from the client side (interface) and returning all the data the interface needs in a single request if the data is requested.
Ex.: If the interface wants to know the metadata of the tokens and account, it will inform on the query, and we will return it on a single query.

On this PR;
- Create a Graphql API that extends fuel-core
- Add metadata loaders for accounts and tokens


LuizAsFight:
@luizstacio I was reading and looks like [Apollo Federation](https://www.apollographql.com/docs/federation) is a good option for combining graphql schemas, but it's a more advanced approach that creates a unified schema by defining relationships between different services and allowing queries across these services.

maybe something we can explore in the future

LuizAsFight (approved):
interesting approach, I tested and it's working for both data from node and extended mocked data

luizstacio:
> @luizstacio I was reading and looks like [Apollo Federation](https://www.apollographql.com/docs/federation) is a good option for combining graphql schemas, but it's a more advanced approach that creates a unified schema by defining relationships between different services and allowing queries across these services.
> 
> maybe something we can explore in the future

The federation works in a different way from the stitching is not far way the distance but is the idea of interdependencies like the one in the description.

PR #37, opened 2023-09-14
feat: add TxCard component and related

pedronauck:
⭐ Closes FRO-471
⭐ Closes FRO-472
⭐ Closes FRO-473

### DIFF
--- a/packages/app/src/systems/Transaction/types.ts
+++ b/packages/app/src/systems/Transaction/types.ts
@@ -1,25 +1,18 @@
-import type { LayerIntent } from '@fuel-ui/css';
+export enum TxTypeEnum {
+  'contract-call' = 'contract-call',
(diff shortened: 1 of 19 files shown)


packages/app/src/systems/Transaction/types.ts
```diff
@@ -0,0 +1,33 @@
+import type { Transaction } from '@fuel-explorer/graphql';
```
luizstacio, line 1: Can we use types from the ts-sdk to avoid unnecessary abstractions? Or from the generated GraphQL endpoint?

pedronauck: This is coming from the `graphql` package being generated by `graphql-codegen` against the schema

luizstacio (approved):

PR #49, opened 2023-09-25
feat: add home page

pedronauck:


luizstacio (changes requested):
- Light theme is not been store; once I refresh the page, the theme changes to dark again
- Between render and font family load the things change place on the UI (we can investigate later if takes too much time)

luizstacio (approved):

PR #52, opened 2023-09-26
feat: add Helm Chart for new infrastructure

liviudm:


helm/fuel-explorer/values.yaml
```diff
@@ -0,0 +1,82 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  repository: ghcr.io/fuellabs/fuel-explorer
+  pullPolicy: Always
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: "af1e196"
```
luizstacio, line 11: Does this tag refer to some deployed images inside the repo?

.github/workflows/helm-publish.yml
```diff
@@ -0,0 +1,38 @@
+name: Build and publish Helm Chart
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    types: [opened, synchronize]
+  release:
+    types: [published]
```
luizstacio, line 10: With this config, will the helm be published on every PR? should we make it just on the main branch? Or would this be capable of deploying preview envs?

luizstacio (changes requested):

helm/fuel-explorer/values.yaml
liviudm: this is just a fallback. What gets deployed is set in the infrastructure repo https://github.com/FuelLabs/fuel-deployment-v2/pull/37/files - please see `kustomize/environments/fuel-dev-2/fuel-apps/dev-fuel-explorer/release.yaml` line 18

.github/workflows/helm-publish.yml
liviudm: this is a good catch, thanks for the feedback. I just copied from your docker build pipeline. We should run this pipeline only if changes are made to `helm/**`. I'll update the rules

luizstacio (approved):

PR #53, opened 2023-09-26
chore: add graphql internally on nextjs

pedronauck:


luizstacio (changes requested):
We should hold on this to evaludate how to do production releases using this infra.

luizstacio (approved):

PR #59, opened 2023-09-28
feat: change tx page layout

luizstacio:


packages/app/src/systems/Transaction/component/TxInfo/TxInfo.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <Card.Header className="gap-0">
+        <Card.Description className="font-medium text-sm h-6 flex items-center">
+          {name}
+        </Card.Description>
+        <Card.Description className="font-medium text-md text-current my-1 h-6 flex items-center color-2">
+          {children}
+        </Card.Description>
+        <Card.Description className="text-sm flex items-center">
+          {description || '  '}
+        </Card.Description>
```
matt-user, line 33:
```suggestion
        {description &&
          <Card.Description className="text-sm flex items-center">
            {description}
          </Card.Description>
        }
```
would this be better?

matt-user (changes requested):

packages/app/src/systems/Transaction/component/TxInfo/TxInfo.tsx
luizstacio: This would make the description space to be smaller the idea of the empty space it to keep the space as it's and not change the size of the card

matt-user (approved):

PR #64, opened 2023-10-18
feat: add TxScript component

pedronauck:
close: #67

Examples:
1. https://fuel-explorer-lxgp17dm6-fuel-labs.vercel.app/tx/0xc8359ad3f0b9f449b6205c28f05fc267df9e5a78e9f707e841920804687a999f
2. https://fuel-explorer-lxgp17dm6-fuel-labs.vercel.app/tx/0x94770af8572b28edbab902978953cc3fef15af28316273b3e1d9260981dd0eac
3. https://fuel-explorer-lxgp17dm6-fuel-labs.vercel.app/tx/0xb6ede856e14254ff7ac53ab52ab3acff52aa4dbf54c939342ed0e56536a902c9
4. https://fuel-explorer-lxgp17dm6-fuel-labs.vercel.app/tx/0x2d873b1f1094520995ef83a860666a72e01761011b31c23b08de57e071c89cc7


packages/graphql/src/services/extender/index.ts
```diff
@@ -7,4 +10,6 @@ export const ExtenderTypeDefs = typeDefs;
 export const ExtenderResolvers = {
   Transaction: TransactionDomain.createResolvers(),
   TransactionConnection: TransactionConnectionDomain.createResolvers(),
+  JSON: GraphQLJSON,
+  JSONObject: GraphQLJSONObject,
```
luizstacio, line 14: I think we don't need this part here, and just use the receipts, if we want to use the operations from the ts-SDK, we should return inside the transaction itself, typing it. We can probably to the type from the ts-sdk also, and just convert from typescript to graphql

luizstacio (changes requested):

luizstacio:
I think the transaction type is not looking good, maybe we could add as a new line, and use the icon, we have on the card... or just a new line. But the two columns look weird.

packages/graphql/src/services/extender/extender.graphql
```diff
@@ -1,16 +1,13 @@
+scalar JSON
+scalar JSONObject
```
luizstacio, line 2: We can remove this scalars now correct?

pedronauck: ✅

luizstacio (approved):

PR #77, opened 2023-10-25
feat: add advanced Transaction screen

LuizAsFight:
Closes FRO-653


packages/app/src/systems/Core/components/ViewMode/ViewMode.tsx
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <Flex
+      align="stretch"
+      justify="center"
+      className="bg-gray-3 p-1 rounded h-9"
+    >
+      <Flex
+        align="center"
+        justify="center"
+        className={`flex-1 rounded px-6 cursor-default ${
```
pedronauck, line 23: I guess the best way to handle these and also to follow the pattern, is using the `tailwind-variants` and you can use `data-[]` to style or `variants`

pedronauck: There's an example here https://github.com/FuelLabs/fuel-explorer/blob/pn/feat/acc-page/packages/app/src/systems/Account/components/AccountTabs/AccountTabs.tsx#L61-L72

packages/app/src/systems/Core/components/JsonViewer/JsonViewer.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export type JsonViewerProps = {
+  data: object | unknown[];
+};
+
+export function JsonViewer({ data, ...props }: JsonViewerProps) {
+  const classes = styles();
+  const ctx = useRadixTheme();
+  return (
+    <JsonView
```
pedronauck, line 18: It's missing just the header of the code editor with the `JSON` title and the `Copy` button as in the layout
https://www.figma.com/file/VTWcuPxsUtKZDk0iKn3y7T/Fuel-Portal-Product-Board?type=design&node-id=2068-72543&mode=design&t=JnqSU5ZVHb0VuptY-4

pedronauck (changes requested):

pedronauck (approved):

PR #84, opened 2023-10-27
feat: contract page -> balance

### LINKED ISSUE #74: Contract Page -> Balances
On the contact page on the balances tab, the user should see all the balances a contract holds. The assets should use the same components as the assets of the user, which means summarize by assetId, in this case, contract balances don't have UTXOs. This way, we can remove the toggle option.

To query the contract balances we can use: `contractBalances`

Design: https://www.figma.com/file/VTWcuPxsUtKZDk0iKn3y7T/Fuel-Portal-Product-Board?type=design&node-id=2068-72575&mode=design&t=QvLFwOte6x138asK-4

LuizAsFight:
Closes #74


pedronauck (changes requested):
CleanShot 2023-11-03 at 19 19 19@2x

isn't getting the Ethereum icon

pedronauck (approved):

PR #85, opened 2023-10-27
chore: improve graphql dev command

pedronauck:
Now after this pull request, in order to have codegen working smoothly just need to run `pnpm dev` inside `packages/graphql` folder, before it was very complicated to get codegen working properly after change any stuff inside the graphql package


packages/graphql/src/bin.ts
```diff
@@ -1,9 +1,8 @@
...
+
 config();
 
 import app from './server';
 
 // Start the server:
-app.listen(4444, () =>
-  console.log('🚀 Explorer api running at http://localhost:4444/graphql'),
-);
+app.listen(4444);
```
luizstacio, line 8: Why remove the log here?

packages/graphql/tsup.config.mjs
```diff
@@ -1,13 +1,20 @@
...
 const graphqlLoaderPlugin = graphqlLoaderPluginPkg.default;
-const defConfig = {
+
+export default defineConfig((options) => ({
   outDir: 'dist',
   splitting: true,
   format: ['esm', 'cjs'],
   sourcemap: true,
-  clean: true,
+  clean: false,
```
luizstacio, line 13: Why not clean the files?

luizstacio (changes requested):

packages/graphql/tsup.config.mjs
pedronauck: because this was generating errors in the watch mode

packages/graphql/src/bin.ts
pedronauck: because I changed it to show just after the codegen

LuizAsFight (approved):

luizstacio (approved):

PR #87, opened 2023-10-28
feat: add account transactions page

### LINKED ISSUE #70: Account Page -> Transactions
The account page transactions tab should show the users all the transactions made by the user, sorted by the newer ones to the old ones. Once the user clicks on the transaction, it should open the transaction page.

Notes;
- Fuel-core returns the transactions from an owner out of order, we can sort the data on the GraphQL side.
- Fuel-core does not support pagination on the transactions by owner we could right now, just return the first 200 transactions.

Design: https://www.figma.com/file/VTWcuPxsUtKZDk0iKn3y7T/Fuel-Portal-Product-Board?type=design&node-id=2068-72496&mode=design&t=QvLFwOte6x138asK-4

pedronauck:
⭐ Closes #70


luizstacio (changes requested):
I think would be good to make min height the same between tabs.

luizstacio (approved):

PR #94, opened 2023-11-01
feat: add block page header and tabs

### LINKED ISSUE #90: Block Page
The block page is a group of tabs with a brief summary of the block. On this task, we should create only the header and the tabs for later adding the contents on the following tasks.

Design: https://www.figma.com/file/VTWcuPxsUtKZDk0iKn3y7T/Fuel-Portal-Product-Board?type=design&node-id=2068-72545&mode=design&t=S2Jr6aSBRekhoQju-4

matt-user:
closes #90

### DIFF
--- a/packages/app/src/systems/Block/screens/BlockScreen.tsx
+++ b/packages/app/src/systems/Block/screens/BlockScreen.tsx
@@ -1,0 +1,35 @@
+'use client';
+
+import type { Maybe } from '@fuel-explorer/graphql';
+import { Flex, Text, VStack } from '@fuels/ui';
+import { IconCube } from '@tabler/icons-react';
+import { useState } from 'react';
+import { PageTitle } from '~/systems/Core/components/PageTitle/PageTitle';
+import {
+  ViewMode,
+  ViewModes,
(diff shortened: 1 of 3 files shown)


packages/app/src/systems/Block/screens/BlockScreen.tsx
```diff
@@ -0,0 +1,35 @@
...
+export function BlockScreen({ blockNumber }: BlockScreenProps) {
+  const [viewMode, setViewMode] = useState<ViewModes>(ViewModes.Simple);
+
+  return (
+    <VStack>
+      <PageTitle icon={<IconCube size={24} stroke={2.4} />}>
+        <Flex justify="between" className="flex-1">
+          <Flex align="center" gap={'5'}>
+            Block
+            <Text size="4">#{blockNumber}</Text>
```
LuizAsFight, line 26: this should match other pages style.

I think you can use <PageSubtitle>. or smth like this

LuizAsFight (changes requested):

pedronauck (approved):

PR #95, opened 2023-11-01
feat: add predicate section for accounts

### LINKED ISSUE #75: Account Page -> Predicate
On the account page predicate tab, the user should be able to see a predicate's bytecode, abi, and signature. We should add a bytecode because we don't have predicate metadata yet.

We should not show the tab predicate if the address is not a predicate.
To know if an address is a predicate, we need to check inside the transaction history of the address and see if an input of type Coin, with the same address, contains a predicate, if yes address is considered a predicate. The bytecode to be shown is this predicate bytecode. This can be done on the GraphQL side and returns only the data in a structured way to the FE side.

Design: https://www.figma.com/file/VTWcuPxsUtKZDk0iKn3y7T/Fuel-Portal-Product-Board?type=design&node-id=2068-72528&mode=design&t=QvLFwOte6x138asK-4

pedronauck:
⭐ Closes #75


packages/app/src/app/account/[id]/transactions/page.tsx
```diff
@@ -15,4 +15,4 @@ export default async function AccountTransactions({
 }
 
 // Revalidate cache every 10 seconds
-export const revalidate = 10;
+export const revalidate = 1;
```
Comment (luizstacio) [inline packages/app/src/app/account/[id]/transactions/page.tsx:18]: Let's keep 10 seconds

packages/app/src/app/transactions/[page]/page.tsx
```diff
@@ -8,10 +8,10 @@ type PageProps = {
...
 
-export const revalidate = 10;
-
 export default async function Transactions({ params: { page } }: PageProps) {
   if (!page || page === '0') return redirect('/transactions/1');
   const txs = await getLastTxs({ page: Number(page) });
   return <TxList transactions={txs} />;
 }
+
+export const revalidate = 1;
```
Comment (luizstacio) [inline packages/app/src/app/transactions/[page]/page.tsx:17]: We can keep 10 seconds here. Also it's possible to add conditional logic here? If the status of the transaction is success or failure the transaction will not change anymore and don't need to be revalidated

packages/app/src/systems/Account/actions/get-predicate-bytecode.ts
```diff
@@ -0,0 +1,27 @@
...
+  const owner = parseAddressParam(input.owner);
+  const { data } = await sdk.getPredicateBytecode({ owner });
+  const nodes = data.transactionsByOwner.nodes;
+  const inputs = nodes.flatMap((n) => n.inputs ?? []);
+  const isPredicate = inputs.some((i) => findBytecodeInput(owner, i));
+  if (!isPredicate) return null;
+  const filtered = inputs.filter((i) => findBytecodeInput(owner, i));
+  const first = filtered[0];
+  if (first.__typename !== 'InputContract') return first.predicate;
+});
```
luizstacio, line 27: This logic should be on the Graphql endpoint, we could create a resolver on the graphql `predicate(address: string)` that does the magic of checking and return the data needed on the UI

luizstacio (changes requested):
- Again consider doing less changes on the overall application when doing features, it makes difficult to review the PR with orthogonal changes.

packages/app/src/app/account/[id]/transactions/page.tsx
Comment (pedronauck) [reply packages/app/src/app/account/[id]/transactions/page.tsx:18]: ✅

packages/app/src/app/transactions/[page]/page.tsx
Comment (pedronauck) [reply packages/app/src/app/transactions/[page]/page.tsx:17]: ✅

packages/app/src/systems/Account/actions/get-predicate-bytecode.ts
pedronauck: ✅

luizstacio (changes requested):
- (This tx)[https://fuel-explorer-11lxmw70s-fuel-labs.vercel.app/tx/0x96e4020a05d97d69f99d0e08008c9b5318ed5edff315065fe5e405b94bb2bc1d] It's showing a predicate tag on a contractCall but the transaction not contains a predicate.
- On predicate endpoint in the graphql we should only return predicate as predicateData can be different for each transaction.

luizstacio (approved):

PR #96, opened 2023-11-02
feat: block page simple

### LINKED ISSUE #91: Block Page -> Transactions
On the block page, transactions tab, the user should see a list of the transactions in the block.

- Pagination is not required as we currently do not have support for pagination on the inner data of a block.
We should query just the minimal info required to display on the list to avoid the query failing, as a block can contain multiple transactions with lots of data.
- When clicking on a Transaction Item, it should open the transaction page.

Design: https://www.figma.com/file/VTWcuPxsUtKZDk0iKn3y7T/Fuel-Portal-Product-Board?type=design&node-id=2529-37680&mode=design&t=S2Jr6aSBRekhoQju-4

matt-user:
closes #91 

should be merged after #90

### DIFF
--- a/packages/app/src/app/account/[id]/page.tsx
+++ b/packages/app/src/app/account/[id]/page.tsx
@@ -3,7 +3,7 @@ import { AccountScreen } from '~/systems/Account/screens/AccountScreen/AccountSc
 
 type AccountProps = {
   params: {
-    id?: string | null;
+    id: string | null;
   };
 };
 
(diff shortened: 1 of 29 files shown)


pedronauck (dismissed):

pedronauck (changes requested):
Can you link the `blockHeight` in the transaction? There's any place with link for blocks pages I guess

luizstacio:
> Can you link the `blockHeight` in the transaction? There's any place with link for blocks pages I guess

We can do this in another task also

matt-user:
> > Can you link the `blockHeight` in the transaction? There's any place with link for blocks pages I guess
> 
> We can do this in another task also

I created a separate issue https://linear.app/fuel-network/issue/FE-9/link-to-block-page

pedronauck (approved):

PR #98, opened 2023-11-06
feat: block page advanced

matt-user:


LuizAsFight:
> @matt-user can you use the JSON preview that we are using on the script component to show the Block JSON instead of plain string?

the component should do it automatically if you pass the right props

LuizAsFight (commented):
the "Transactions" tab is out of pattern with other screens, should be the same tab. this one can be done in other PR as it's orthogonal to this PR objective

luizstacio (approved):
@matt-user can you use the JSON preview that we are using on the script component to show the Block JSON instead of plain string?

PR #101, opened 2023-11-07
feat: block page mobile

matt-user:
closes FE-3


luizstacio (changes requested):
As we spoke on the sync today, we need to fix the break point for the intermediary screens also.

luizstacio (approved):

PR #107, opened 2023-11-10
refactor: remove type change on click

matt-user:
closes FE-27


luizstacio (approved):

luizstacio (changes requested):
We should add to this links a hover with color primary

luizstacio:
Can you also add the hover on the block?
Screenshot 2023-11-13 at 2 09 56 PM

luizstacio:
- When we are rendering a asset_id we don't need to have a link;
Screenshot 2023-11-13 at 2 10 25 PM

luizstacio:
> Can you also add the hover on the block? Screenshot 2023-11-13 at 2 09 56 PM

Can you do this also?

LuizAsFight (changes requested):
can we make the address always link style, instead of only when hovering?

on this way the user knows it's a link, otherwise user needs to hover to know it's a link

LuizAsFight (changes requested):
image

add pointer to copy button

LuizAsFight (changes requested):
- add link style also when not hovering
- remove useless icon

image

LuizAsFight (changes requested):
before:
image

----
after
image

---
address size is looking way smaller than before, and also disaligned

LuizAsFight (changes requested):
image

not showing sender / recipient anymore

luizstacio (changes requested):
The size and font weight should be the same of previous versions. It looks a bit bigger and bold now.
Screenshot 2023-11-15 at 8 12 32 PM
Screenshot 2023-11-15 at 8 12 19 PM

luizstacio (dismissed):

packages/ui/src/components/Copyable/Copyable.tsx
```diff
@@ -51,6 +51,7 @@ export const Copyable = createComponent<CopyableProps, 'span'>({
         {children}
         <Tooltip content={tooltipMessage}>
           <IconButton
+            className="cursor-pointer"
```
LuizAsFight, line 54: this fix would be better on UI side ?

`<IconButton>` is a Button I'd expect it to be pointer by default, unless it's disabled/readOnly

LuizAsFight: this would also fix all other IconButtons

image

LuizAsFight (dismissed):

luizstacio (approved):

PR #109, opened 2023-11-11
fix: utxo item component

LuizAsFight:
Closes FE-34

Ref FE-33


matt-user (commented):
The Utxo component still doesn't stack right on certain breakpoints.

md:

Screenshot 2023-11-13 at 12 53 45 PM

sm:

Screenshot 2023-11-13 at 12 52 24 PM

LuizAsFight:
> The Utxo component still doesn't stack right on certain breakpoints.


the problem is that transaction screen is not responsive as a whole... cards should be breaking.

if UTXO component takes the whole view, it will show correctly

matt-user (approved):

PR #110, opened 2023-11-13
feat: code block shows json preview

matt-user:
closes FE-20


luizstacio (changes requested):
This is not related to fix itself but as the component is the same, Can you change the arrow direction in the show more link, it should be inverted.

Screenshot 2023-11-14 at 10 15 22 PM

luizstacio (approved):

PR #112, opened 2023-11-13
feat: make entire component clickable

matt-user:
closes FE-21
closes FE-22


pedronauck (changes requested):
If you just add an `onClick` in the entire element without verifying the action when triggering other actions, it will open the collapse; also, you need to check in the `onClick` if the click is just in the collapsible component


https://github.com/FuelLabs/fuel-explorer/assets/2029172/f8a348a7-d265-492b-8b17-53f7bcb2beb5

luizstacio (changes requested):
Clicks on this area should not collapse the item also:
Screenshot 2023-11-15 at 7 46 24 PM

luizstacio (approved):

PR #114, opened 2023-11-14
fix: performance improvement on graphql

pedronauck:
⭐ Closes FE-27


luizstacio (changes requested):
The clicks are faster but not good enough the user has a delay. Maybe we should do something like what we do on the home page that has a loader on the content this would be possible? How much effort?

luizstacio (approved):

PR #115, opened 2023-11-14
fix: format units according to asset decimals + unknown asset

LuizAsFight:
Closes FE-31

this PR address fully the format of amounts being showed across the explorer. will format to assets decimal config OR format the raw amount if it’s a unknown asset (unknown asset decimals)
- update to `fuels/assets@0.1.3`  to get new asset utils functions
- share assetId with components that didn’t have it
- apply correct amount formatted to this pages/components
  - BalanceItem
  - Utxos
  - TxAssetItem
  - TxInput
  - TxOutput


luizstacio (changes requested):
The amount format is great, but it should not have any decimal... as a decimal `.0` for integer means one more 0 that don't exist.

Account example: https://fuel-explorer-j0r8fpcif-fuel-labs.vercel.app/account/fuel18e7amfxs60urq7h97xhdsa3rnykpcn0valkxsjfkjcrh2xqmyvpq4ay9jn It should be showing `1,052,399,877` only.

luizstacio (approved):

PR #117, opened 2023-11-14
feat: improve search

matt-user:
closes FE-12


luizstacio (changes requested):
We can improve a little bit more by creating just a single resolver name `search(query: string)`.

packages/graphql/src/domains/Search.ts
```diff
@@ -0,0 +1,121 @@
...
+      block: {
+        id: string;
+        header: {
+          height: string;
+        };
+      };
+      transaction: {
+        id: string;
+      };
+    };
```
luizstacio, line 68:
```suggestion
    /** @todo: Get types from the query directly instead of creating custom types */
    type Result = {
      transactionsByOwner: {
        nodes: {
          id: string;
        }[];
      };
      contract: {
        id: string;
      };
      block: {
        id: string;
        header: {
          height: string;
        };
      };
      transaction: {
        id: string;
      };
    };
```

luizstacio (approved):

PR #119, opened 2023-11-15
feat: improve search UI

matt-user:
closes FE-13


luizstacio (changes requested):
Screenshot 2023-11-20 at 10 09 41 PM

1. ~~You can remove the placeholder.~~
2. ~~The helper text could be smaller and need to have a bit more spacing on the top of it.~~

Screenshot 2023-11-20 at 10 10 54 PM

4. The error should be shown in the place of the helper text with red letter's.

Screenshot 2023-11-20 at 10 12 34 PM

6. ~~The dropdown should have the same width of the input.~~
7. ~~The search should be trigger with enter also.~~

Screenshot 2023-11-20 at 10 13 56 PM

9. ~~The Focus when using keyboard should have the same background color.~~
10. When loading the actions should become a spinner and the dropdown should be closed, right now, it opens with previous data and them change with the new data.

packages/app/src/app/transactions/page.tsx
```diff
@@ -1,17 +1,33 @@
...
   };
 };
 
 export default async function Transactions({
-  searchParams: { page = '1' },
+  searchParams: { page = '1', searchQuery = '' },
 }: PageProps) {
   const txs = await getLastTxs({ page: Number(page) });
-  return <TxList transactions={txs} page={page} />;
+  const searchResult = await search({ query: searchQuery });
```
pedronauck, line 19: Hm, I'm pretty sure this is not the best way to handle this. In fact, you don't need to fetch in the server component, since this is a mutation that just happens after some interaction...

The best approach for this is:
- Create a server action using the `'use server'` that will handle the fetch logic
- Use `useAct`, or `form`, or `startTransaction` to trigger after click on submit

So, by doing this you will avoid all this prop dilling and workaround back-and-fourth of data in the components tree in order to retrieve the data.

pedronauck (changes requested):

pedronauck (changes requested):
CleanShot 2023-11-28 at 15 19 56@2x

- Need to adjust the dropdown to work in less resolutions
- When clicking on submit it's adding the loading in both buttons `clear` and `submit`, should be just in the `submit`

pedronauck (approved):

luizstacio (changes requested):
- On Enter it should trigger the search.
- The error should not be highlight with hover, and letters should be red.

luizstacio (approved):

PR #122, opened 2023-11-17
feat: make transaction list inline

LuizAsFight:
Closes FE-15
Closes FE-58

- improve logic of gas calculation
- add logic to calculate network Fee of tx
- make transaction list cards to have inline information and one per line
- make Block page use component `<TxList>`
- add transaction fee to transaction screen and list
- format gas fee as `0.00002 ETH` 
- format gas used, gas fee and gas limit as integer `123,123,123`
- remove Skeleton components from TxCard and TxList
- created new `<TransactionWrapper` component:

```
<LoadingWrapper
  isLoading={isLoading}
  loadingEl={<LoadingBox className="w-32 h-6" />}
  regularEl={tx.time?.fromNow}
/>
```


luizstacio (changes requested):
I didn't found a storybook to see the loader state of the new transaction line.

luizstacio (approved):

PR #123, opened 2023-11-20
feat: nav search

matt-user:
closes FE-14


matt-user:
> functional review:
> 
> on my first access on the preview link this is what I see:
> 
> * I see the two icons (clean and green check), they should show only if there's something in the input field
> 
> image

I received feedback from @luizstacio to always display the search input icon buttons.  @luizstacio was this feedback for both search inputs or just for the nav search?

luizstacio:
> > functional review:
> > on my first access on the preview link this is what I see:
> > 
> > * I see the two icons (clean and green check), they should show only if there's something in the input field
> > 
> > image
> 
> I received feedback from @luizstacio to always display the search input icon buttons. @luizstacio was this feedback for both search inputs or just for the nav search?

The idea is to show the buttons only on the navbar on the header it should behave as before.

LuizAsFight (changes requested):
in mobile, it's almost going over the fuel Icon, needs to have some space

image

LuizAsFight (changes requested):
search animation need some improvement, my 2 cents:

- could be rtl
- a bit faster
- 
cc
@luizstacio

LuizAsFight (changes requested):
I also think the UX can improve , rn you need 2 clicks to close the search after typing something, which can be confusing.

for that we can:
- open the input animated rtl + change the search icon to a close button, will stay in some position. if user wants to open/close search bar, just click in same place

closed:
image


opened (should also include theme switch):
image


cc @luizstacio

LuizAsFight (approved):

PR #127, opened 2023-11-23
feat: polish all loading pages

LuizAsFight:
Closes FE-45
Closes FE-57


luizstacio (changes requested):
The loader on the
Account Page -> Transaction: is showing the same from the assets page.

luizstacio (approved):

PR #128, opened 2023-11-23
feat: add error/notfound pages

pedronauck:
⭐ Closes FE-47


luizstacio:
@pedronauck I tried to just access a random URL but it does not show the 404 page. Do we need some configuration?

luizstacio (changes requested):
After accessing a error page, and clicking on the button back to home, the theme selector bugs and the theme becomes light with the button bugged without state.

Screenshot 2023-11-28 at 3 43 05 PM

luizstacio (approved):

PR #133, opened 2023-11-29
feat: add cursors where applicable

matt-user:
closes FE-40

### DIFF
--- a/packages/ui/src/components/IconButton/IconButton.tsx
+++ b/packages/ui/src/components/IconButton/IconButton.tsx
@@ -4,6 +4,7 @@ import { getIconSize } from '../../hooks/useIconProps';
 import { useVariants } from '../../hooks/useVariants';
 import type { Variant, WithVariants } from '../../hooks/useVariants';
 import { createComponent } from '../../utils/component';
+import { cx } from '../../utils/css';
 import type { Colors, PropsOf } from '../../utils/types';
 import { Icon } from '../Icon/Icon';
 import type { IconContext } from '../Icon/useIconContext';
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/IconButton/IconButton.tsx
```diff
@@ -48,12 +47,15 @@ export const IconButton = createComponent<IconButtonProps, 'button'>({
...
+        className={
+          !isDisabled && !props['aria-readonly'] ? 'cursor-pointer' : ''
+        }
       >
         {isLoading ? (
           <Spinner color="current" size={getIconSize(size, iconSize)} />
         ) : (
           <Icon
-            className={cx(iconClassName, 'cursor-pointer')}
+            className={iconClassName}
```
LuizAsFight, line 58: you need to keep the `cx(` here right?

```suggestion
            className={cx(iconClassName)}
```

LuizAsFight (changes requested):
when hovering the iconButtons they're having this green background, which doesn't exist in master. I suspect the reason is the change I requested

image

packages/ui/src/components/IconButton/IconButton.tsx
matt-user: no I don't think we need cx here.  If you look [here](https://github.com/FuelLabs/fuel-explorer/commit/dd96475bb42c8c302582d641d70ac78e658ac2bc#diff-142661a2119190194a9fe74c7977c95c8bf76d7db987d8051984701b2d53fe06) iconClassName was not initially wrapped with cx.  I believe cx is used when we want to add classes to an existing class.

LuizAsFight (approved):

PR #134, opened 2023-11-29
feat: align eth icon

matt-user:
closes FE-52


packages/app/src/systems/Asset/components/AssetItem/AssetItem.tsx
```diff
@@ -22,12 +30,14 @@ export function AssetItem({
   return (
     <HStack {...props} align="center">
       {asset?.icon ? (
-        <Image
-          src={asset.icon as string}
-          width={ICON_SIZE}
-          height={ICON_SIZE}
-          alt={asset.name}
-        />
+        <Flex className="w-[40px] h-[40px] items-center justify-center">
```
LuizAsFight, line 33: is it possible to achieve same result using `w-9` or any other number width ?

it's better as it will match with `TxIcon` sizing as well

LuizAsFight (changes requested):

LuizAsFight (approved):
nice!

PR #139, opened 2023-12-08
chore: update beta-5

pedronauck:


deployment/Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM node:20-slim AS base
 
 # Expose the ENVs to the env of the container
 ENV PORT="${PORT}"
-ENV FUEL_PROVIDER_URL="${FUEL_PROVIDER_URL}"
+ENV FUEL_PROVIDER_BETA5="${FUEL_PROVIDER_BETA5}"
```
luizstacio, line 7: Why we change the name of the provider here?

luizstacio (changes requested):

packages/app/src/systems/Transaction/component/TxScreen/TxScreenSimple.tsx
```diff
@@ -132,19 +132,13 @@ export function TxScreenSimple({ transaction: tx, isLoading }: TxScreenProps) {
...
                   <LoadingBox className="w-28 h-5 mt-1" />
                 </>
               }
-              regularEl={
-                <>
-                  Gas used: {formatZeroUnits(tx.gasUsed || '')}
-                  <br />
-                  Gas limit: {formatZeroUnits(tx.gasLimit || '')}
-                </>
-              }
```
luizstacio, line 147: Gas limit is available for transaction of type Script, you can find it on the policies array.

deployment/Dockerfile
LuizAsFight: `FUEL_PROVIDER_URL` was better IMO to avoid needing to change later.

LuizAsFight (changes requested):
in transaction screen, can we hide the fee block when it's a mint transaction?

PR #140, opened 2023-12-12
feat: improve loading speed of pages

luizstacio:
Improve first paint, on pages to avoid long waiting times, and improve loaders in all pages using next.

For now, the discovers are;
- `useRouter`  is slower than <Link />  as the component link is optimized to prefetch the pages. In this way, the navigation is available right after the page is rendered. With useRouter, it has a delay between the page is loaded and actions can be taken;
- As lower number of files is better for the performance. Creating a single file page with dynamic routing is faster than having multiple files;
- Common components need to be rendered on the layout to optimize for partial render;
- Using Suspense is better than using the loading file;
- Refactor routes to use a common enum `Routes` to improve manutenability;
- Refactor files and structures to follow the same standard;
- Add cache keys on next.js `dynamic`, `fetchCache` and `invalidations` .

On this PR;
- [x] Home page
- [x] Navbar
- [x] Recent Transactions
- [x] Transaction page
- [x] Contract page
- [x] Block page
- [x] Account page 

closes FE-71


pedronauck (commented):
I was navigating through the application, and when I clicked on some accounts this happened:

CleanShot 2023-12-20 at 11 20 08@2x

It was a very random navigation I did

pedronauck:
After I cleaned up the application cache for your link, this didn't happen anymore and the page loading was absurdly fast.

https://github.com/FuelLabs/fuel-explorer/assets/2029172/04f16d16-0002-4001-ac05-ac116bb5e1d1

luizstacio:
> I was navigating through the application, and when I clicked on some accounts this happened:
> 
> CleanShot 2023-12-20 at 11 20 08@2x
> 
> It was a very random navigation I did

I think is a problem. On the API getting into the request limit

pedronauck (approved):

LuizAsFight (approved):

LuizAsFight:
very nice work around here !

PR #143, opened 2024-01-16
fix: network fee

matt-user:
closes FE-55


matt-user:
I forgot to screenshot, but before this change we were using the minFee which does not account for the fee from gas spent.  The tx I was using before the change was 0.000000879 and after the change it was 0.0000008887 which matches the wallet.

luizstacio (approved):

LuizAsFight (approved):

PR #144, opened 2024-01-21
feat: improve search navbar

LuizAsFight:
Closes FE-86


luizstacio (changes requested):
- Once we click in one of the links in the search, we could close the dropdown.
- As we have the search all the time in the header, in the home page we could hide to avoid displaying two search bars.
- On mobile I think it would be better to have the search as a popup with just a icon on the header. Once you click the icon the popup opens with fullscreen. Right now using on mobile is very not good as is not possible to see the values correctly.

luizstacio (approved):
Request are going to be addressed on another PR

luizstacio (approved):

PR #147, opened 2024-01-29
feat: navbar mobile

LuizAsFight:
Closes FE-93


pnpm-lock.yaml
matt-user, line 1: Should this file be modified if we did not change any of our deps?

matt-user (commented):
looks good, I tested a couple of different phone screens.  Just a comment about the lock file.

pnpm-lock.yaml
LuizAsFight: good point, I did some cleans due to beta-5 upgrade, but should not have commited. thanks

matt-user (approved):

PR #148, opened 2024-01-30
fix: themes not returning correctly

matt-user:
closes FE-87


matt-user:
for some reason `updateThemeAppearanceClass` does not update `useThemeContext` so `useThemeContext` would always return dark.

pedronauck (changes requested):
When reloading using network with no cache and fast 3g it's happening this...

CleanShot 2024-02-01 at 19 44 34@2x

luizstacio (changes requested):
With the new implementation the theme configuration is lost on every refresh.

matt-user:
> With the new implementation the theme configuration is lost on every refresh.

@luizstacio if you are referring to the tx screen this happens on main too.  I'll investigate the issue and fix it in this pr.

luizstacio:
> > With the new implementation the theme configuration is lost on every refresh.
> 
> @luizstacio if you are referring to the tx screen this happens on main too. I'll investigate the issue and fix it in this pr.

@matt-user yeah this is true, I didn't notice that before. Last merge this PR and fix in another task.

luizstacio (approved):

pedronauck (approved):

PR #155, opened 2024-02-02
feat: add tx mint custom layout

pedronauck:
⭐ Closes FE-84


luizstacio (changes requested):
- I think we should keep the same design of the other TX pages, but remove the empty things and add just the Outputs Section. It looks a bit weird with all things as blocks.
- And add blocks for this other information on the side.
Screenshot 2024-02-01 at 22 08 22

luizstacio (changes requested):
With the new change the script transaction is broken: https://fuel-explorer-c1guco7wj-fuel-labs.vercel.app/tx/0x546ad80a1244efcc41b3e208d0231b4d1c284b0fe776f69a731435cd9a706a86/simple
Screenshot 2024-02-05 at 19 42 00

1. I think on the mint tx, we could change the title from Outputs to `Minted Assets`.
2. We could use the same layout for the outputs, as we have the asset, the to: in this case is the `contract` and the other infos could be bellow it.
Screenshot 2024-02-05 at 19 43 13

luizstacio (changes requested):
https://fuel-explorer-b470hi2yc-fuel-labs.vercel.app/tx/0xbb6101d8ba1cfc84d04552f484ed6d27844a8f9ef11357bd1c8904bf4f35ddd9/simple
Screenshot 2024-02-12 at 14 23 21

luizstacio (approved):

PR #169, opened 2024-02-09
feat: make theme configuration work cross applications

luizstacio:
On this PR;
- Add a force component that setTheme if undefined to ensure the theme is on localStorage
- Once theme on storage changes both application receive the new theme
- Remove the need for next-themes inside `@fuels/ui`

### DIFF
--- a/packages/app-explorer/src/systems/Core/components/TopNav/TopNav.tsx
+++ b/packages/app-explorer/src/systems/Core/components/TopNav/TopNav.tsx
@@ -7,9 +7,9 @@ import { useEffect, useState } from 'react';
 import { SearchWidget } from '../Search/SearchWidget';
 
 export function TopNav() {
+  const { theme, setTheme } = useTheme();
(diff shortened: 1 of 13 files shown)


packages/app-explorer/src/systems/Core/components/TopNav/TopNav.tsx
```diff
@@ -58,6 +58,7 @@ export function TopNav() {
   const themeToggle = (
     <Nav.ThemeToggle
       whenOpened="no-effect"
+      theme={theme!}
```
matt-user, line 61: should we have a default theme to fall back on in the case theme is null?

luizstacio: It should not have a undefined state as the ThemDefault forces the state to be either dark or light.

matt-user (approved):

LuizAsFight (approved):

PR #188, opened 2024-02-22
chore: add warning update on bridge

pedronauck:


luizstacio (changes requested):
I installed the Fuel Wallet 0.13.11 via zip and the popup not show up.

packages/app-portal/src/systems/Settings/providers/FuelConnectProvider.tsx
```diff
@@ -19,7 +19,7 @@ export function FuelConnectProvider({ children }: ProvidersProps) {
       theme={theme}
       fuelConfig={{
         devMode: IS_DEV,
-        connnectors: defaultConnectors(),
+        connectors: defaultConnectors(),
```
luizstacio, line 22: We need to remove Fuelet for now to import the connectors use `FuelWalletConnector and FuelWalletDevelopmenConnector` here.

luizstacio (changes requested):

luizstacio (approved):

PR #196, opened 2024-02-26
refactor: improve how we build classNames

helciofranco:
We've been running `twMerge` and `clsx` on every re-render, it led to some kind of performance issue.
So this PR is replacing `twMerge` with a simple `clsx` and also running it **ONLY** when it is really necessary.

With these changes, we can get a great performance on rendering.

| ⚠️ Before | 🚀  After |
|----------|----------|
| Screenshot 2024-02-26 at 14 58 56 | Screenshot 2024-02-26 at 14 59 42 |

Reference: [prefer twJoin over twMerge](https://github.com/dcastil/tailwind-merge/blob/v2.2.1/docs/when-and-how-to-use-it.md#joining-internal-classes)


packages/ui/src/components/List/List.tsx
```diff
@@ -16,8 +16,8 @@ export type ListItemProps = PropsOf<'li'>;
 
 export const ListRoot = createComponent<ListProps, 'ul'>({
   id: 'List',
-  className: ({ className, icon, type = 'none' }) => {
-    return styles({ type, withIcon: Boolean(icon) }).root({ className });
```
pedronauck, line 20: why did you remove the className from here?

packages/ui/src/utils/component.tsx
```diff
@@ -34,9 +35,22 @@ export function createComponent<
...
-    const baseClass =
-      typeof getClass === 'function' ? getClass(props as P) : getClass;
-    const classes = cx(baseClass, className, fClass(id));
+    const baseClass = useMemo<string | undefined>(() => {
+      if (typeof getClass === 'function') {
+        return getClass(props);
+      }
+
+      return getClass;
+    }, [props]);
```
pedronauck, line 44:
```suggestion
    const baseClass = useMemo(() => {
      return typeof getClass === 'function' ? getClass(props) : getClass
    }, [props]);
```

pedronauck (changes requested):

packages/ui/src/utils/component.tsx
```diff
@@ -33,11 +34,16 @@ export function createComponent<
...
-    const classes = cx(baseClass, className, fClass(id));
-    const itemProps = { ref, className: classes, ...props } as any;
+  const Comp = forwardRef<T, P>((props, ref) => {
+    const className = useMemo<string | undefined>(() => {
+      if (typeof getClass === 'function') {
+        return cx(getClass(props), fClass(id));
+      }
+
+      return cx(getClass, props.className, fClass(id));
+    }, [props.className]);
```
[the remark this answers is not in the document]
helciofranco: as we aligned on slack:
https://github.com/FuelLabs/fuel-explorer/pull/196/commits/134da2548212989d879686c6c290fc49b41a5c08

pedronauck (approved):

PR #202, opened 2024-02-27
fix: nextjs error logs

helciofranco:
This PR fixes how `Next.JS` resolves `app-portal` imports.

https://github.com/FuelLabs/edge/issues/102

| No more errors |
|----------|
| clean logs |


packages/app-explorer/package.json
```diff
@@ -17,10 +17,10 @@
     "@fuel-explorer/graphql": "workspace:*",
     "@fuel-ts/math": "0.74.0",
     "@fuel-wallet/sdk": "0.15.2",
-    "@fuels/assets": "0.1.5",
+    "@fuels/assets": "0.15.2",
     "@fuels/ui": "workspace:*",
     "@tabler/icons-react": "2.46.0",
-    "@tanstack/react-query": "5.17.19",
+    "@tanstack/react-query": "4.36.1",
```
helciofranco, line 23: This downgrade is intentional.

I did it just to keep only one version of `react-query` in our final bundle.
Mostly because everywhere was on `v4` and we're not using any `v5 feature` currently.

downgrade

We might upgrade it in a follow-up PR.

packages/app-explorer/src/app/bridge/history/page.tsx
```diff
@@ -1,8 +1,10 @@
+'use client';
```
pedronauck, line 1: this page shouldn't have `use client` here

packages/app-explorer/src/app/bridge/page.tsx
```diff
@@ -1,8 +1,10 @@
+'use client';
```
pedronauck, line 1: this page shouldn't have `use client` here

pedronauck (changes requested):

packages/app-explorer/src/app/bridge/history/page.tsx
helciofranco: do you remember the reason? 
there are a few hooks in this page blocking the build due to the server components.

helciofranco:
```
app-explorer:build: ../app-portal/src/systems/Assets/containers/AssetsDialog.tsx
app-explorer:build: Error: 
app-explorer:build:   × You're importing a component that needs useState. It only works in a Client Component but none of its parents are marked with "use client", so they're Server Components by default.
```

packages/app-explorer/src/app/bridge/page.tsx
helciofranco: https://github.com/FuelLabs/fuel-explorer/pull/202/commits/68500623b8aba8f718e7a2105f60a834b26a7b1b

pedronauck (approved):

PR #237, opened 2024-03-14
docs: updated project list and status

Rubyt0x:


.vscode/settings.json
```diff
@@ -9,5 +9,8 @@
   },
   "tailwindCSS.experimental.classRegex": [
     ["tv\\(([^)]*)\\)", "[\"'`]([^\"'`]*).*?[\"'`]"]
-  ]
+  ],
+  "[json]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  }
```
helciofranco, line 15: can we revert this change? we're using `biomejs` to format our files

packages/app-portal/src/systems/Ecosystem/data/projects.json
```diff
@@ -1,434 +1,420 @@
...
+    "name": "Acumen",
     "url": "https://acumen.network/",
-    "tags": ["DeFi", "Lending", "Real World Assets"],
+    "tags": ["DeFi", "Real World Assets"],
     "description": "Acumen is bridging DeFi and Real-world assets.",
     "github": "https://github.com/AcumenProtocol",
     "twitter": "https://twitter.com/acumenofficial",
     "discord": "https://discord.gg/vnUPngfB5D",
-    "image": "acumen",
-    "status": ["Building"]
```
helciofranco, line 12: I've noticed that `status` is currently required, we need to make it optional in order to fix the build.

https://github.com/FuelLabs/fuel-explorer/blob/f98e8b5a81da5b74be613cbdfe334deb25c3d76f/packages/app-portal/src/systems/Ecosystem/types.ts#L10

helciofranco (approved):

PR #247, opened 2024-03-22
fix: query key

helciofranco:
- Create a `useEthBalance` to reduce re-renders since just only place in the bridge should watch the balance.
- Invalidate ETH balance whenever the block number changes (wagmi doc recommendation).

#### Reference
[Example: Watching a Users' Balance](https://wagmi.sh/react/guides/tanstack-query#example-watching-a-users-balance)


LuizAsFight (approved):
nice !

PR #257, opened 2024-03-26
fix: show correct theme inside connect wallet popup

### LINKED ISSUE #238: Provide the correct theme for the Connectors UI
Currently the Connector UI is always render on light theme instead of track the theme of the application. The Connector Provider, has the ability to receive a theme prop to ensure it can render on light or dark theme.

Screenshot 2024-03-14 at 22 03 14

LuizAsFight:
Closes #238


HoTandy (approved):
Perfect!

helciofranco (approved):

PR #291, opened 2024-04-04
feat: add L1 indexer to fetch contract logs for the bridge

helciofranco:
Add L1 indexer, it'll sync contract events logs from a given start block number.
After that, it'll keep watching the blockchain for new logs and blocks.

New env vars have been added to the graphql:
- `FUEL_CHAIN_NAME` (example `fuelBeta5`)
- `ETH_CHAIN_NAME` (example `sepolia`)
- `ETH_ALCHEMY_ID` (your alchemy key)
- `ETH_INFURA_ID` (your infura key – used as fallback to alchemy)
- `ETH_INITIAL_BLOCK` (block number that got our contracts)

###  📷  Indexer demo
| `pnpm sync:bridge` |
| --- |
| <video src="https://github.com/FuelLabs/fuel-explorer/assets/7074983/8780375a-b352-48b4-9975-bc1f6093b9ab" /> |

### 💾  API demo
| Query | Result |
| --- | --- |
| Screenshot 2024-04-04 at 18 01 16 | Screenshot 2024-04-04 at 18 01 27 |


.github/workflows/docker-publish.yml
```diff
@@ -50,4 +50,4 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
           image: ghcr.io/fuellabs/fuel-explorer
           dockerfile: deployment/Dockerfile.v2
-          context: ./packages/graphql-new
+          context: .
```
luizstacio, line 53: We should not change the root dir here; this makes the image publish all the code from the repo instead of only the graphql package, drastically increasing the image size.

luizstacio (changes requested):

packages/graphql-new/src/application/uc/WatchBridgeContractLogs.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    console.log(`\n👀 Watching logs from #${blockNumber.number} block`);
+
+    // Sync the logs
+    try {
+      await this.watchLogs();
+    } catch (e) {
+      console.error(e);
+      throw new Error('Failed to watch logs');
+    }
```
luizstacio, line 37: How watchLogs work? If the socket is closed as example by the server it autoReconnects?

.github/workflows/docker-publish.yml
helciofranco: @luizstacio indeed :( 

I'll need to figure out how to get rid of this – let me find a time to check that today.

I've added this because `graphql-new` package has a peer dependency on the `@fuel-explorer/contract-ids`.
(Outside that `context`)

Does that make sense?

helciofranco:
Closing this as it has become too outdated. 
We can reopen it later if still relevant.

Issue #251, opened 2024-03-26
Info message link in bridge opens in the same browser tab


HoTandy:
The link doesn't open in a new tab, it navigates away from the bridge and opens the github link on the same tab.

`security in our `[docs](https://github.com/FuelLabs/fuel-bridge/blob/main/docs/ARCHITECTURE.md)

![image](https://github.com/FuelLabs/fuel-explorer/assets/114662397/73da0255-213f-4f99-ba5b-3dfb76147625)

Change this so the link is opened in a different tab and the user is not navigated away from the bridge.

PR #305, opened 2024-04-20
fix: docs link should open in a new tab

Manav2808:
Closes: [Info message link in bridge opens in the same browser tab #251](https://github.com/FuelLabs/fuel-explorer/issues/251)


packages/app-portal/src/systems/Bridge/pages/Bridge.tsx
```diff
@@ -122,12 +121,13 @@ export const Bridge = () => {
           Any assets deposited to Fuel can take up to {timeToWithdrawFormatted}{' '}
           to withdraw back to Ethereum. Learn more about our architecture and
           security in our&nbsp;
-          <Link
-            isExternal
+          <a
```
helciofranco, line 124: We just need change to keep using that `Link`.
It adds a pretty icon and other styles to it.

You can keep `target` and `rel` as you did. That looks good.

```suggestion
          <Link
```

helciofranco (changes requested):
Hey there! 

Thanks for you contribution, left just a small comment.

packages/app-portal/src/systems/Bridge/pages/Bridge.tsx
Manav2808: Done.

packages/app-portal/src/systems/Bridge/pages/Bridge.tsx
```diff
@@ -123,8 +123,9 @@ export const Bridge = () => {
           to withdraw back to Ethereum. Learn more about our architecture and
           security in our&nbsp;
           <Link
-            isExternal
+            target="_blank"
```
helciofranco, line 126: You missed the `isExternal` prop (that property is adding an icon there)

```suggestion
            isExternal
            target="_blank"
```

helciofranco (changes requested):

packages/app-portal/src/systems/Bridge/pages/Bridge.tsx
Manav2808: Oh, my bad. Now Done.

PR #315, opened 2024-04-23
feat: ComboBox component

arthurgeron:
Created Auto Complete component
autocomplete


packages/ui/src/components/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,250 @@
...
+    <div className={className} onClick={onClick} onKeyDown={onKeyDown}>
+      <Text>{itemNameSelector?.(suggestion) ?? (suggestion as string)}</Text>
+    </div>
+  );
+}
+
+export const ComboBoxItem = memo(
+  createComponent({
+    id: 'ComboBoxItem',
+    baseElement: ComboBoxItemBase,
```
pedronauck, line 213: it's a better approach use the render prop directly instead of using this `baseElement`, the idea of the `baseElement` is when you are using polymorphic components, and you want to define a *base* element that will respond as or when you are just using an external library and you would like to use some external component as base.

when you are creating components from scratch - as you are doing here - you should use the render property instead of `baseElement`, this will avoid also you created this `*Base` components functions and using `memo`, etc.

packages/ui/src/components/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,250 @@
...
+  type KeyboardEvent,
+  type MouseEvent,
+} from 'react';
+import { createComponent, withNamespace } from '../../utils/component';
+import { Flex } from '../Box';
+import { Input, type InputFieldProps } from '../Input';
+import { Popover } from '../Popover';
+import { Text } from '../Text';
+
+export interface ComboBoxProps<T = string> {
```
pedronauck, line 17: It's a nice thing use the `BaseProps` and `PropsOf` types helper to define base elements props as well and not just the custom ones (look at some other components like `Nav` they have example of this usage)

packages/ui/src/components/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,250 @@
...
+      }
+      case 'Escape':
+        setIsPopoverOpen(false);
+        break;
+      default:
+        break;
+    }
+  };
+
+  const childWithProps = useMemo(() => {
```
pedronauck, line 150: Instead of passing use this would be better to use a context, you can check inside the `Collapsible` component that it has a example of usage

packages/ui/src/components/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,250 @@
...
+  onItemSelected,
+  inputFieldProps,
+  children,
+}: ComboBoxProps<T>) {
+  const [isPopoverOpen, setIsPopoverOpen] = useState(false);
+  const [itemSelected, setItemSelected] = useState(false);
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [filteredSuggestions, setFilteredSuggestions] =
+    useState<Array<T>>(suggestions);
+  const debounceTimeout = useRef<NodeJS.Timeout>();
```
pedronauck, line 74: This debounce logic could be improved also, it's kinda too much imperative, would be nice to have it inside a hooks, or even better you can use [useDebounce](https://github.com/streamich/react-use/blob/master/docs/useDebounce.md) hook from `react-use` (we are already using it as lib here)

packages/ui/src/components/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,250 @@
...
+  };
+
+  const onBlur = (e: FocusEvent<HTMLInputElement, Element>) => {
+    _onBlur?.(e);
+    if (strict && !itemSelected) {
+      onChange(null);
+    }
+  };
+
+  const onKeyDown = (event: KeyboardEvent<HTMLInputElement>) => {
```
pedronauck, line 133: react-use already has a [useKeyPressEvent](https://github.com/streamich/react-use/blob/master/docs/useKeyPressEvent.md) also if you want to avoid need to implement this by yourself (I know it's a simple stuff, but just a matter of trying to follow what was already being used in the project)

pedronauck (changes requested):

packages/ui/src/components/ComboBox/ComboBox.tsx
arthurgeron: @pedronauck  Thanks for the feedback! I just have a couple observations.

I used `base` component for clarity, not to indicate polymorphism, as I had to keep type inference with generics working properly after applying `memo` to the component (I wonder if type inference with generics would work out of the box with that util?).

I have a doubt about the `memo` part, the idea is that the list could be exponentially big( I was thinking about using "Virtualized Lists" in the future; This'd scale badly when combined with the usage of Context API in the `ComboBoxContent` component, as all renders, even without changes to actua list data, would trigger a new render into the Content component and consequently leak renders into all the list items. That means that we could get to a scenario where, with a 0 debounce, the entire list of X items would re-render for each letter typed in the search box, which would generate noticeable hiccups given a big enough set of data.

How can optimizations integrate well with the `createComponent` util? Does it apply, allow or integrate well with optimizations such as `memo`?

packages/ui/src/components/ComboBox/ComboBox.tsx
arthurgeron: My idea of using injection was to allow for devs to use their own custom components, if they accept at least the required props, for more flexibility, is that not needed/used in our codebase? If it's, we'd need to also export the context so that devs can create their own custom components to achieve the same flexibility

pedronauck: Got it, yeah but isn't something we would like to offer, flexibility isn't the thing here since this design system it will be used more for us than external and as I mentioned, the idea is having more standard and without frictions as possible. When you start to give flexibility, you opened a door for more complexity, as well more friction.

packages/ui/src/components/ComboBox/ComboBox.tsx
arthurgeron: useDebounce [depends](https://github.com/streamich/react-use/blob/ade8d3905f544305515d010737b4ae604cc51024/src/useDebounce.ts#L13) on React's life cycle, since it uses `useEffect` it's going to run after renders.
The logic of this component is event based, non-dependant on react's lifecycle (hence why you can use it without passing value/onChange), which is faster, especially for typing/filtering fields.

I could change the entire logic, but it'd make the list visually slower, as it'd need to wait for a render + timeout and then it'd filter the list (which takes O(n) time). It'd also make the use of value/onChange obligatory.

I can create a hook for a event-based, or instead let me know what else you belive could improve this logic.

Thank you!

packages/ui/src/components/ComboBox/types.ts
```diff
@@ -0,0 +1,42 @@
+import type { BoxProps, FlexProps } from '../Box';
```
pedronauck, line 1: Instead of having this as a separated file, you can ad its content inside the `ComboBox.tsx` file

packages/ui/src/components/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,275 @@
...
+      value,
+      onItemSelected,
+      inputRef,
+    } = props as ComboBoxProps<string>;
+
+    const [isPopoverOpen, setIsPopoverOpen] = useState(false);
+    const [itemSelected, setItemSelected] = useState(false);
+    const [filteredSuggestions, setFilteredSuggestions] =
+      useState<Array<string>>(suggestions);
+    const debounceTimeout = useRef<NodeJS.Timeout>();
```
pedronauck, line 59: What about the `useDebounce` from `react-use` was not possible to use it here?

packages/ui/src/components/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,275 @@
...
+  baseElement: Input,
+});
+
+export const ComboBoxTrigger = createComponent<{}, typeof Popover.Trigger>({
+  id: 'ComboBoxTrigger',
+  baseElement: Popover.Trigger,
+  render: (Root, { ...props }) => {
+    return (
+      <Root {...props}>
+        <div />
```
pedronauck, line 170: Beautiful workaround 🤣

pedronauck (changes requested):

pedronauck (approved):

PR #316, opened 2024-04-25
fix: casting of contracts and transactions as any

arthurgeron:
Replaced `any` castings by the correct usages, focusing on contracts and transactions.


helciofranco (approved):

packages/app-explorer/src/systems/Transaction/component/TxScripts/TxScripts.tsx
```diff
@@ -358,8 +357,7 @@ function parseJson(
   if (!item) return {};
   return Object.entries(item).reduce((acc, [key, value]) => {
     if (!value || key === '__typename') return acc;
-    if (typeof value === 'object')
-      return { ...acc, [key]: parseJson(value as any) };
+    if (typeof value === 'object') return { ...acc, [key]: parseJson(value) };
```
pedronauck, line 360:
```suggestion
    if (typeof value === 'object') {
      return { ...acc, [key]: parseJson(value) }
    }
```

pedronauck (changes requested):

packages/app-explorer/src/systems/Transaction/component/TxScreen/TxScreenLoader.tsx
```diff
@@ -2,5 +2,5 @@ import { TxScreenSimple } from './TxScreenSimple';
 
 export function TxScreenLoader() {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  return <TxScreenSimple isLoading transaction={{} as any} />;
+  return <TxScreenSimple isLoading transaction={undefined} />;
```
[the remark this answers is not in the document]
arthurgeron: I've changed the typing for TxScreenSimple so we can conditionally permit for a transaction object to not be passed.

packages/app-explorer/src/systems/Transaction/component/TxScripts/TxScripts.tsx
arthurgeron: Good catch! Fixed

pedronauck (approved):

PR #318, opened 2024-04-26
feat: sort ecosystem projects by "live" flag

helciofranco:
- Projects marked as `live` are sorted first in the projects list.
- Then, a secondary alphabetical sorting is applied.

| Example |
| --- |
| Screenshot 2024-04-26 at 14 22 36 |


packages/app-portal/src/systems/Ecosystem/hooks/useEcosystem.tsx
```diff
@@ -1,24 +1,39 @@
...
     if (filter) {
-      return projects.filter((project) => {
+      const filtered = projects.filter((project) => {
         return project.tags.includes(filter);
       });
+      return filtered.sort(sortProjects);
     }
 
-    return projects;
+    return projects.sort(sortProjects);
```
arthurgeron, line 36: This might be a way to centralize the logic into a single flow,  but it's not necessarily a better approach, just a suggestion
```JS
    const { projects = [], search, filter } = state.context;

    let processedProjects = projects;

    if (search || filter) {
      processedProjects = processedProjects.reduce((acc, project) => {
        const matchesSearch = search ? project.name.toLowerCase().includes(search.toLowerCase()) : true;
        const matchesFilter = filter ? project.tags.includes(filter) : true;

        if (matchesSearch && matchesFilter) {
          acc.push(project);
        }

        return acc;
        }, []);
      }
    
    return processedProjects.sort(sortProjects);
```

arthurgeron (approved):

packages/app-portal/src/systems/Ecosystem/hooks/useEcosystem.tsx
helciofranco: indeed, great suggestion, let's check later if we need to support filter + search at same time.
currently the state machine is allowing only one or other – so I'm not sure if it was an existing product requirement.

LuizAsFight (changes requested):
codewise looking good but preview PR link is broken

helciofranco:
> codewise looking good but preview PR link is broken

@LuizAsFight  I updated our env. vars to use `beta-5` API again, now it's all set.

HoTandy (approved):
Looks good!

helciofranco:
@LuizAsFight can you review it again plz?

HoTandy (approved):
I've done a quick second review and all looks good.

LuizAsFight (approved):

HoTandy (commented):
Now all DApps 👍

PR #353, opened 2024-06-03
fix: swaylend broken page and add e2e to validate all projects

helciofranco:
- Fix `Swaylend` broken page (replace with the proper URL)
- Add e2e test to verify if all projects are working as expected.


packages/app-portal/src/systems/Core/utils/url.ts
```diff
@@ -4,7 +4,12 @@ const trimPath = (path = '') => path.replace(trimRegex, '');
 
 export const getUrlHostName = (url: string) => {
   try {
-    return new URL(url || 'https://#').hostname;
+    const { hostname } = new URL(url || 'https://#');
+    if (hostname.startsWith('www.')) {
+      return hostname.replace('www.', '');
+    }
+
+    return hostname;
```
arthurgeron, line 12: Looks good but it can be shortened
```suggestion
    return hostname.replace(/^www\./,'');
```

arthurgeron (approved):

packages/app-portal/src/systems/Core/utils/url.ts
helciofranco: Thanks, I'll keep it in my notes for a following PR. 
Auto-merge was enabled.

PR #383, opened 2024-06-20
fix: timing on bridge bugged in first minute

rodrigobranas:
Screenshot 2024-06-20 at 11 18 34
Screenshot 2024-06-20 at 11 15 00


packages/app-commons/src/utils/dayjs.ts
```diff
@@ -0,0 +1,29 @@
+import dayjs from 'dayjs';
+import englishLocale from 'dayjs/locale/en.js';
+import relativeTime from 'dayjs/plugin/relativeTime';
+
+export function createDayjs() {
+  dayjs.extend(relativeTime, {});
+  const localeObject = {
+    ...englishLocale,
+    name: 'fuel',
+    relativeTime: {
```
rodrigobranas, line 10: I had to keep all the properties here, otherwise I breaks que date conversion

LuizAsFight (approved):

PR #386, opened 2024-06-20
feat: update design system colors

arthurgeron:
- Added back primary and secondary text color variables
- Updated all colors to match the new Design System colors
- Standardized borders to use `border` color, and set it to be `gray-6`, which is aligned with Radix's color science.
- Fixes constant declared inside ViewMode, which is usually used in client components, being imported by server components.
- Fixes Bridge Skeleton/Loading state not matching new size of elements
- Fixes label contrast color in brand/green accent buttons, now set to #111111
- Added prefetch to Transaction ViewMode links

image
image
image


LuizAsFight (changes requested):
transaction and block page not accessible, internal server error 500

arthurgeron:
> transaction and block page not accessible, internal server error 500

Next.JS still executes some client-component code during SSR or SSG, even for client components, and `router.prefetch` called in `ViewMode` accesses `window` internally, which causes the error in prod. 
I've fixed the code to only run client side.

LuizAsFight (approved):

LuizAsFight (approved):

PR #440, opened 2024-08-10
Fix: Creating new operations parser

rodrigobranas:
Current operations parser is showing wrong operations and in some cases cashes due to recursive loop, which causes the pod to restart in the cluster.

Following the transactions that causes the crash:

0x3825126eb65b9cc9e62a04fd8b9906ce2b9b4da931c2917c0736832463054f26
0x559115f5d83fa2c327e0dad823d8f2f26b736dbf07032d9050bbbb76ca045e4e
0x5327db1ff455c09fb1f6123c542be23e3a5088b9a2a6cab7a69ac59d8026dbb8
0xb14b1b275a211d9b191d21b36abf1624b833b4c743e3791bbf9fdcc08db98a99
0x75f2e905adcc0cfe8f857acf0c64ff8076441db4656172cd86e85d384b8c005c
0x98cccd4e3c4cda5d4513a460da951a8e02020a86ba1973d1a2f016c0377ee5a9
0xdb4bd0147dabca36c2840601b64fed2c6314f651f25b2dcc4a8d54690c0c8ebb
0xbefe5e1c6fa033dc74d0692d0aecfd36a14337bbb98184c8d1d5d75e8197ea18
0xde7d06e51464fcfc5a213e63471f38314657afcba5dd351336e43b64ced8099a

Also, there are many transactions showing more operations that actually exists:

0xa3d923c0ebbe55670ba875e59534f7ea49ab83f8915b349db9c6dd5ce642559c

In this case there are only 13 receipts, however is showing 19 operations.

That is just one example, there are many others.

With the new algorithm the operations are rendered without consuming almost any memory and in a few milliseconds.

https://github.com/user-attachments/assets/8ce3d883-299c-43ed-92bc-9d2588f6183a


rodrigobranas:
Should we show operations just for Success status? Or Failure should also be showed?

Screenshot 2024-08-11 at 11 22 04

This transaction ran out of gas after the first call, raise a panic and returned.

https://app.fuel.network/tx/0x66d1cc4e21c648d7ae46187d71cc79a8a6f343817cbab322f1205a25da1f2de2/simple

Should be show?

LuizAsFight:
> Should we show operations just for Success status? Or Failure should also be showed?
> 
> The current code prevents showing operations for failure status
> 
> Screenshot 2024-08-11 at 11 22 04
> 
> Following there is an example, the transaction below ran out of gas after the first call, raise a panic and returned. However, the operations were not showed.
> 
> https://app.fuel.network/tx/0x66d1cc4e21c648d7ae46187d71cc79a8a6f343817cbab322f1205a25da1f2de2/simple
> 
> Should I keep this behavior?

can you show how it would look without this filter? I think we should show operations even for fail transactions

LuizAsFight (changes requested):
can we revert pnpm-lock change as there was no change to pnpm packages?

rodrigobranas:
> can we revert pnpm-lock change as there was no change to pnpm packages?

sure, I reverted pnpm-lock to the main branch

rodrigobranas:
> > Should we show operations just for Success status? Or Failure should also be showed?
> > The current code prevents showing operations for failure status
> > Screenshot 2024-08-11 at 11 22 04
> > Following there is an example, the transaction below ran out of gas after the first call, raise a panic and returned. However, the operations were not showed.
> > https://app.fuel.network/tx/0x66d1cc4e21c648d7ae46187d71cc79a8a6f343817cbab322f1205a25da1f2de2/simple
> > Should I keep this behavior?
> 
> can you show how it would look without this filter? I think we should show operations even for fail transactions

transaction: 0x66d1cc4e21c648d7ae46187d71cc79a8a6f343817cbab322f1205a25da1f2de2

Screenshot 2024-08-12 at 10 55 54

LuizAsFight (changes requested):
1- change indentation size to half of what it is now
2- show all indentation size, instead of limiting on 3

LuizAsFight:
Closing in favor of https://github.com/FuelLabs/fuel-explorer/pull/375

PR #449, opened 2024-08-22
feat: add analytics on bridge

luizstacio:


packages/app-portal/src/systems/Bridge/machines/bridgeMachine.tsx
```diff
@@ -148,8 +149,18 @@ export const bridgeMachine = createMachine(
...
+          const isWithdraw = isFuelChain(input.fromNetwork);
+          trackEvent({
+            eventType: 'bridge',
+            eventName: `bridge-${isWithdraw ? 'withdraw' : 'deposit'}`,
+            parameters: {
+              asset: input.asset?.symbol,
+              addressFrom: input.ethAddress,
+              addressTo: input.fuelAddress?.toB256(),
+            },
+          });
```
LuizAsFight, line 163: it's better to track the event inside of the service. reasoning is that there's no guarantee that the bridge process really occurred here.

but as this is a test feature, let's go on this way

LuizAsFight (approved):

Issue #454, opened 2024-08-26
Predicate tab not disabled for accounts


LuizAsFight:
the predicate tab should be enabled only for predicate accounts

PR #458, opened 2024-08-28
fix: predicate tab not disabled for accounts

xlzy905:
Issuse: https://github.com/FuelLabs/fuel-explorer/issues/454

Problem: The account will return a '0x' string
image

PR #482, opened 2024-09-09
fix: preventing 0x predicates to being showed and persisted

rodrigobranas:
- Closes: #454 

Screenshot 2024-09-09 at 11 30 30
Screenshot 2024-09-09 at 11 30 44
Screenshot 2024-09-09 at 11 30 20

### DIFF
--- a/packages/graphql/src/application/uc/NewAddBlockRange.ts
+++ b/packages/graphql/src/application/uc/NewAddBlockRange.ts
@@ -140,7 +140,7 @@ export default class NewAddBlockRange {
   getPredicate(input: GQLInput) {
     if (!['InputCoin', 'InputMessage'].includes(input.__typename)) return;
     const bytecode = (input as GQLInputCoin | GQLInputMessage).predicate;
-    // if bytecode === 0x return;
+    if (bytecode === '0x') return;
     let address = '';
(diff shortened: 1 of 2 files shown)


helciofranco (approved):

arthurgeron (changes requested):
I think E2E tests expected predicate tabs to always show, they probably need to be adjusted

arthurgeron:
This PR overlaps with #458. Though imho adjusting directly in graphql resolve seems like a better solution.

rodrigobranas:
> I think E2E tests expected predicate tabs to always show, they probably need to be adjusted

actually there are different issues, I was indexing 0x predicates and this PR is preventing that for next blocks. For the existing blocks I'm just avoiding return 0x predicates as it's not valid and makes sense while I don't delete them from database.

About the E2E tests, I did not change the behavior of the frontend, just not returning when there is no predicate (considering 0x).

rodrigobranas:
> This PR overlaps with #458. Though imho adjusting directly in graphql resolve seems like a better solution.

I think that does not make sense to handle that on the frontend, it needs to rely on the backend in order to know whether there is or there isn't predicate for the account

arthurgeron (approved):

PR #507, opened 2024-09-13
Account Data Indexing

raghukapur9:
- Added accounts table
- Functions to add/update account entry


packages/graphql/src/domain/Account/AccountModel.ts
```diff
@@ -0,0 +1,9 @@
+import { integer, jsonb, pgTable, serial, text } from 'drizzle-orm/pg-core';
```
rodrigobranas, line 1: we are not using Drizzle anymore, don't need to create the mapping classes

PR #550, opened 2024-09-18
fix: add some deps as external

LeoCourbassier:
- Without this config, importing @fuels/ui was breaking

### DIFF
--- a/packages/ui/tsup.config.mjs
+++ b/packages/ui/tsup.config.mjs
@@ -1,9 +1,9 @@
-import tsconfig from "./tsconfig.json";
+import tsconfig from './tsconfig.json';
 
 const defConfig = {
-  outDir: "dist",
+  outDir: 'dist',
   splitting: true,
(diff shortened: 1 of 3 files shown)


arthurgeron (dismissed):

packages/ui/tsup.config.mjs
```diff
@@ -1,9 +1,9 @@
-import tsconfig from './tsconfig.json';
+import tsconfig from "./tsconfig.json";
```
LuizAsFight, line 1: what formatter you are using? can we revert this format change from single ' to double "? keep only the change needed on external

LuizAsFight (changes requested):

packages/ui/tsup.config.mjs
LeoCourbassier: I'm using biome, I think it may have misconfigured. Yes, I'll fix that

LuizAsFight (approved):

PR #703, opened 2026-09-07
fix(graphql): L1 indexer backoff never resolved and froze every contract cursor

nelitow:
## Summary

The mainnet L1 indexer stopped on 2026-09-03 and stayed silent for four days. The retry backoff in the block timestamp fetch awaited a promise that can never resolve, so a single transient RPC error froze every contract cursor without crashing the process. Users whose withdrawals were claimed on L1 after that point kept seeing them as ready to process and got a nonce already used revert when they retried. This makes the backoff a real sleep, keeps one failing contract from blocking the others, and exits the process when a sync stalls so pm2 restarts it.

## Changes

| Area | Change |
|------|--------|
| L1 indexer retry backoff | Replace the never-resolving promise with a real sleep between getBlock retries |
| L1 indexer stall watchdog | Exit the process when a contract sync runs longer than 10 minutes so pm2 restarts it with the cursor intact |
| L1 indexer error isolation | Log and retry a contract that throws on the next tick instead of killing the loop for every contract |
| L1 indexer tests | Cover the retry path, per-contract error isolation, and the stall watchdog |

### DIFF
--- a/packages/graphql/src/application/uc/IndexL1/IndexL1.test.ts
+++ b/packages/graphql/src/application/uc/IndexL1/IndexL1.test.ts
@@ -127,6 +127,108 @@ describe('IndexL1', () => {
     expect(updateCall?.[1]).toEqual([finalizedNumber + 1, 42]);
   });
 
+  it('retries a failed block timestamp fetch and finishes the cycle', async () => {
+    const log = {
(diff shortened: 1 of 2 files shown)


nelitow:
👍

PR #704, opened 2026-09-07
fix(app-staking): show finalize errors in the withdraw status dialog

nelitow:
## Summary

When finalizing a staking withdrawal reverted on L1, the dialog showed nothing. The error was stored in the machine but the selectors ignored it and checked a state that no transition ever reached. This surfaced last week when users retried an already claimed withdrawal and got a nonce already used revert with no feedback. The dialog now shows the red error banner for finalize failures, clears it when the user retries, and the dead state is gone.

## Changes

| Area | Change |
|------|--------|
| Withdraw status dialog error banner | Include finalize errors in the error selectors, same as the undelegate machine |
| Withdraw retry | Clear the previous finalize error when the user clicks Finalize again so a successful retry does not stay marked as Error |
| Withdraw machine | Remove the unreachable finalizingError state |


nelitow:
👍

PR #706, opened 2026-09-10
chore: build the deploy images in CI and keep the api-lite test run green

nelitow:
## Summary

Nothing built the two images Railway deploys. They are built during `railway up`, the e2e workflow only uses Docker for the local fuel-core chain, and the image-publishing workflow that was removed built a different, now-deleted Dockerfile. A broken Dockerfile would therefore first be discovered during a deploy. A new workflow builds both images on pull requests that touch their inputs and publishes nothing.

Separately, removing the committed block fixtures left api-lite's parity suite registering zero tests. Jest fails any suite with no tests, so the package's test command exited non-zero on a clean checkout even though every real test passed.

## Changes

| Area | Change |
|------|--------|
| CI | New `build-images.yml` builds the api-lite and explorer images on PRs touching `docker/vps/**`, `packages/**` or the lockfiles, and on demand |
| api-lite tests | Parity suite registers a skipped placeholder when no block fixtures are present, instead of failing the whole run |

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -1,0 +1,51 @@
+name: "Build deploy images"
+
+# Railway deploys with `railway up`, which builds docker/vps/Dockerfile.api-lite
+# and docker/vps/Dockerfile.explorer at deploy time. Nothing else builds them:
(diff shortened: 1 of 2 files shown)


nelitow:
👍

PR #707, opened 2026-09-10
feat(api-lite): read blocks from a public bucket-root S3 endpoint

nelitow:
## Summary

Cloudflare R2's public `r2.dev` host addresses a single bucket and serves keys at its root, so using it needs no bucket name, no region and no credentials. The S3 client could not express that shape: path-style addressing inserts the bucket a second time, and virtual-host style would prefix a bucket onto a host that already is one, so every read 404s. An endpoint given without a bucket is now fetched directly as `<endpoint>/<key>`, which makes an anonymous public bucket a valid block source.

## Changes

| Area | Change |
|------|--------|
| S3 fetcher | An endpoint without a bucket fetches `<endpoint>/<key>` over plain HTTPS; a 404 is a missing block, any other status is an error |
| Config | `S3_BUCKET` and `AWS_REGION` are no longer required when `S3_ENDPOINT` is set on its own |
| Tests | Cover key placement on a bucket-root endpoint, and the 404 versus error split |

The authenticated bucket plus region path, including when an endpoint is supplied for the S3 API, is unchanged.

Verified end to end against the mainnet block archive: `BLOCK_SOURCE=s3` with the public endpoint serves mainnet at the chain tip (`servedTip` 63866938 against a chain tip of 63866946, zero index gaps).


nelitow:
👍

nelitow:
:+1:

PR #709, opened 2026-09-11
feat(api-lite): fall back to the node for a height the archive lacks

nelitow:
## Summary

`BLOCK_SOURCE` selects exactly one source, so a height the archive does not have became a gap the index could never advance past. That is the common case at the tip: the recorder publishes a block shortly after it is produced, so a read can miss it. The bucket is complete behind the tip, which means this only bites where the explorer most needs to be current.

The block store can now fall back to the node for a height the archive reports as absent, giving the lookup order memory, disk cache, archive, then node.

## Changes

| Area | Change |
|------|--------|
| Block store | New optional `fallback`, consulted only when the source reports the height absent |
| Block store | A read that fails for any other reason still throws, so an archive problem surfaces as itself rather than silently moving reads to the node |
| Block store | Fallbacks are summarised at most once a minute, with a count, instead of one line per block |
| Startup | When S3 is the source, the RPC source is built anyway and passed as the fallback, so that traffic stays inside the same per-second budget |
| Tests | Cover the fallback being used and indexed like any other block, its absence, a non-miss error not invoking it, and the summarised logging |

## Verification

Ran against an S3 endpoint that answers `404` to every request, so every read had to come from the fallback. `servedTip` advanced in step with the chain tip (`72956401 → 72956426` against a tip of `72956441`) with zero index gaps, and the log carried one summary line rather than one per block. Without the fallback the same setup records every height as a gap.


nelitow:
:+1:

PR #710, opened 2026-09-11
fix(api-lite): bound gunzip output for archive and disk-cache reads

nelitow:
## Summary

A crafted or corrupt gzip object — small on the wire, huge once decompressed — could exhaust the service's 768 MB heap, because neither the S3/R2 archive read path nor the on-disk block cache read path capped how much decompressed data `gunzipSync` was allowed to produce. Both paths now pass a 64 MiB output limit to `gunzipSync`. A too-large archive object is now a named error (not mistaken for a missing block, which would otherwise get silently skipped), and a too-large disk cache file is treated the same way the code already treats a corrupt one — it falls through to the legacy path and then to a cache miss.

## Changes

| Area | Change |
|---|---|
| `packages/api-lite/src/s3/S3BlockSource.ts` | Added `MAX_BLOCK_BYTES` (64 MiB) constant; `fetchRaw` passes it as `maxOutputLength` to `gunzipSync` and rethrows Node's `ERR_BUFFER_TOO_LARGE` as a descriptive `Error` naming the height and limit, distinct from `BlockNotFound` |
| `packages/api-lite/src/store/BlockStore.ts` | `readDisk` passes `MAX_BLOCK_BYTES` as `maxOutputLength` to `gunzipSync`; the existing catch-and-fall-through already treats the resulting error like any other corrupt cache file |
| `packages/api-lite/src/s3/S3BlockSource.test.ts` | New test asserting an over-limit gzip object rejects with an error naming the height and limit, and is not `BlockNotFound` |
| `packages/api-lite/src/store/BlockStore.test.ts` | New test asserting an over-limit disk cache file is handled like a corrupt file (falls through to the source, no throw) |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/api-lite/src/s3/S3BlockSource.test.ts
+++ b/packages/api-lite/src/s3/S3BlockSource.test.ts
@@ -1,5 +1,10 @@
 import { gzipSync } from 'node:zlib';
-import { BlockNotFound, S3BlockSource, createS3Fetcher } from './S3BlockSource';
+import {
+  BlockNotFound,
+  MAX_BLOCK_BYTES,
+  S3BlockSource,
+  createS3Fetcher,
+} from './S3BlockSource';
 
 describe('S3BlockSource', () => {
   it('gunzips gzip objects', async () => {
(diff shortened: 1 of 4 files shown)


nelitow:
:+1:

PR #711, opened 2026-09-11
fix(api-lite): validate staking page size and cursor params

nelitow:
## Summary

The staking REST pagination parameters accepted invalid values that could
degrade or crash a request. A negative `last` (page size) was not rejected
and reached SQLite as an unlimited `LIMIT`, causing a single request to
process every matching row. A non-numeric `after`/`before` cursor silently
became `NaN` instead of being rejected. Both now throw a clear validation
error instead of reaching the database.

## Changes

| Area | Change |
| --- | --- |
| `packages/api-lite/src/staking/PaginatedParams.ts` | Reject `last` values that are 0, negative, non-integer or NaN with a new `ValidationError` message; keep the existing over-cap message and the default of 10 when absent. Reject non-integer `after`/`before` cursors with a `ValidationError` instead of letting them become `NaN`. |
| `packages/api-lite/src/staking/PaginatedParams.test.ts` | Add coverage for `last` of -1, 0, 1.5 and "abc", and for non-numeric `after`/`before` cursors. |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/api-lite/src/staking/PaginatedParams.test.ts
+++ b/packages/api-lite/src/staking/PaginatedParams.test.ts
@@ -38,4 +38,46 @@ describe('PaginatedParams', () => {
       'Maximum page size allowed is 4',
     );
   });
+
+  // SQLite treats a negative LIMIT as unlimited, so last=-1 must not reach
+  // L1Index's `LIMIT @limit`.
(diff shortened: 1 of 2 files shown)


nelitow:
:+1:

PR #712, opened 2026-09-11
fix(api-lite): give the s3-mode rpc fallback its own rate limit, separate from rpc-mode streaming

nelitow:
## Summary

In s3 mode, `BlockStore`'s rpc fallback (used only for heights the S3
archive doesn't have yet, mostly a few blocks at the tip) shared one
5-blocks-per-second bucket with rpc mode's full block stream. That capped
catch-up after downtime in s3 mode at 5 blocks/s even though the fallback's
actual load is much lighter than a full stream. Added a separate,
higher-default rate limit for the fallback so s3-mode catch-up isn't
throttled by a limit sized for rpc mode's continuous streaming.

## Changes

| Area | Change |
| --- | --- |
| `packages/api-lite/src/config.ts` | Add `RPC_FALLBACK_MAX_BLOCKS_PER_SECOND` (integer, default 20), same validation style as `RPC_MAX_BLOCKS_PER_SECOND`, mapped to `rpcFallbackMaxBlocksPerSecond`. |
| `packages/api-lite/src/config.test.ts` | Add coverage for the default (20) and an invalid value. |
| `packages/api-lite/src/main.ts` | Build `rpcSource` with `cfg.rpcMaxBlocksPerSecond` in rpc mode and `cfg.rpcFallbackMaxBlocksPerSecond` in s3 mode; comment explains why the two differ. |
| `docker/vps/deploy.md` | Document `RPC_FALLBACK_MAX_BLOCKS_PER_SECOND` next to `RPC_MAX_BLOCKS_PER_SECOND` in the Railway `api` variables block. |
| `docker/vps/.env.example` | Document `RPC_FALLBACK_MAX_BLOCKS_PER_SECOND=20` next to `RPC_MAX_BLOCKS_PER_SECOND` (this file's `BLOCK_SOURCE=s3` is the deployment where the new variable is actually live). |


nelitow:
:+1:

nelitow:
:+1:

PR #713, opened 2026-09-11
fix(api-lite): serve /health during boot before fuel-core answers chain params

nelitow:
## Summary

Railway's healthcheck had nothing to poll while api-lite waited on
fuel-core's chain params at boot, because the HTTP port wasn't opened until
after that wait (up to 30 seconds of backoff) completed. The server now
starts listening immediately with a minimal handler that reports booting
status, then swaps in the real request handler once the app is fully built,
without ever closing or reopening the port.

## Changes

| Area | Change |
| --- | --- |
| `packages/api-lite/src/bootServer.ts` | New module: an `http.Server` with a mutable, swappable request handler, defaulting to a 503 "booting" handler. |
| `packages/api-lite/src/bootServer.test.ts` | New test: 503 booting responses before swap, real handler responses after swap. |
| `packages/api-lite/src/main.ts` | Starts the boot server on `cfg.port` before `chainParamsWithRetry`; swaps in the real app's request listener once `createApp` returns; shutdown closes the boot server instead of the app's never-listening server. |
| `packages/api-lite/src/server.test.ts` | New test wiring `createApp`'s server listener into `createBootServer` the same way `main.ts` does, verifying `/health` responds correctly end to end. |


nelitow:
:+1:

nelitow:
:+1:

PR #714, opened 2026-09-11
fix(api-lite): fetch the L1 finalized block once per poll tick instead of once per contract

nelitow:
## Summary

Every poll tick, the L1 poller was asking the RPC node for the finalized
block number separately for each of the 7 tracked contracts, even though
that number is the same for all of them within one tick. Moved the fetch up
to run once per tick and reused the result for every contract, cutting 7
identical RPC calls down to 1. If that single fetch fails, the whole tick is
skipped and logged once, instead of failing (and logging) once per contract.

## Changes

| Area | Change |
| --- | --- |
| `packages/api-lite/src/l1/L1Poller.ts` | `tick()` fetches the finalized block once and passes it to `syncContract()` as a new parameter; `syncContract()` no longer fetches it itself. A failed fetch now skips the whole tick and logs once, rather than skipping and logging per contract. |
| `packages/api-lite/src/l1/L1Poller.test.ts` | Updated direct `syncContract()` calls for the new signature; fixed a test whose only assertion (finalized-fetch call count) stopped proving "every contract ran" once fetches became shared; added tests for the once-per-tick fetch with 3 mocked contracts and for the whole-tick-skip-on-failure behavior. |


nelitow:
:+1:

nelitow:
:+1:

PR #715, opened 2026-09-11
fix(app-explorer): handle nginx 429 responses without retry storms, add an error boundary

nelitow:
## Summary

nginx answers rate-limited requests with a 429 and an HTML body. The explorer's `api.ts` called `res.json()` on every non-ok response, so a 429 threw a `SyntaxError` instead of a meaningful error, and React Query's default retry (3 attempts, no backoff) piled more requests into the same rate-limit burst. Separately, there was no `ErrorBoundary` anywhere in `app-explorer`, so any uncaught render error produced a blank white screen. `api.ts` now reads the error body safely and throws a typed `ApiError` (with a fixed message for 429s, and a parsed-or-statusText message otherwise); the `QueryClient` no longer retries 4xx errors and backs off exponentially (1 s, 2 s) for the rest; and `<App />` is now wrapped in a minimal `ErrorBoundary` with a "Reload" button.

## Changes

| Area | Change |
|---|---|
| `packages/app-explorer/src/systems/Core/utils/api.ts` | Added `ApiError` (status + message); non-ok responses now read the body as text and only `JSON.parse` it, instead of blindly calling `res.json()`; 429 always throws a fixed "Too many requests, please slow down" message without touching the (HTML) body |
| `packages/app-explorer/src/main.tsx` | `QueryClient` default `retry` now returns `false` for any 4xx `ApiError` (429 included) and otherwise retries up to 2 times; `retryDelay` is exponential starting at 1 s; `<App />` wrapped in the new `ErrorBoundary`; `refetchOnWindowFocus` default left untouched (already overridden per-query in the two hooks that need it) |
| `packages/app-explorer/src/systems/Core/components/ErrorBoundary/ErrorBoundary.tsx` | New class component catching render errors from its subtree, showing a minimal message and a "Reload" button (`@fuels/ui` `Button`) |
| `packages/app-explorer/src/systems/Core/utils/api.test.ts` | New unit tests for the 429/HTML, 500/JSON, and non-JSON-non-429 error paths (see report: this package's Jest config cannot currently be parsed, pre-existing on `main`, so these have not been executed) |


nelitow:
:+1:

nelitow:
:+1:

PR #716, opened 2026-09-11
fix(api-lite): cap GraphQL batching at 10 and add a 15-level query depth limit

nelitow:
- `batching: true` is now `batching: { limit: 10 }`, made explicit instead of relying on graphql-yoga's default for the boolean shorthand.
- Added a hand-written depth-limit validation rule (`packages/api-lite/src/graphql/depthLimit.ts`) rejecting operations nested deeper than 15 levels, since `@escape.tech/graphql-armor-max-depth` isn't installed in this workspace. 15 was chosen (over the original 12) to keep headroom above the explorer's deepest real query, `transactionDetails` (depth 11).
- `maxTokens` was investigated and skipped: graphql-yoga 5.6.0 (the installed version) has no `maxTokens` option or `useMaxTokens` plugin.
- Introspection and rate limiting are untouched.
- Tests: `packages/api-lite/src/server.test.ts` — batch of 11 operations rejected, a query nested 16 levels deep rejected with `Query depth limit of 15 exceeded`, a query nested exactly 15 levels deep is not rejected for depth, a normal explorer query still succeeds.


nelitow:
:+1:

nelitow:
:+1:

PR #717, opened 2026-09-11
fix(api-lite): verify archived blocks against fuel-core's block id

nelitow:
## Summary

Blocks served from the S3/R2 archive had their consensus signature patched
in from fuel-core, but the archive's own block id was never checked against
fuel-core's id for that height. A compromised or corrupted archive could
have served fabricated block content with no way to detect it. The
signature query to fuel-core now also fetches the block id for each height
(same single call, no added cost), and a block is rejected with a clear
error if the two ids don't match. Heights fuel-core hasn't caught up to yet
are left unchecked, same as before.

## Changes

| Area | Change |
| --- | --- |
| `packages/api-lite/src/fuelcore/FuelCoreClient.ts` | `blockSignatures` now selects `id` alongside `consensus` and returns `Map<number, { id, signature }>` instead of `Map<number, string>`. |
| `packages/api-lite/src/graphql/resolvers/blocks.ts` | `withSignatures` compares fuel-core's `id` for each height against the archived block's own `id`; on a mismatch it logs and throws a `GraphQLError`, refusing to serve that block. A missing entry (node lagging) still skips the check, unchanged from before. |
| `packages/api-lite/src/graphql/blockSignatures.test.ts` | Added matching-id and mismatched-id cases; updated existing mocks to the new `{ id, signature }` shape. |
| `packages/api-lite/src/fuelcore/FuelCoreClient.test.ts` | Updated `blockSignatures` tests for the `id` field in the query and result. |
| `packages/api-lite/src/graphql/resolvers.test.ts` | Updated the shared `blockSignatures` mock to the new `{ id, signature }` shape. |


nelitow:
:+1:

nelitow:
:+1:

PR #718, opened 2026-09-11
perf(api-lite): serve the home dashboard from an edge-cached GET /dashboard snapshot

nelitow:
**Summary:** The home page's recent-blocks table polls `getBlocksDashboard` over GraphQL every 5 seconds. This adds a REST `GET /dashboard` route backed by the same 5s in-process cache the resolver already used, fronted by an nginx edge cache with the same 5s TTL, so repeated polls under load hit nginx instead of re-running the resolver's block-range read. The frontend now tries `/api/dashboard` first and falls back to the original GraphQL query on any failure, the same fallback shape `useHomeCharts` already uses for `/api/charts`.

| File | Change |
|---|---|
| `packages/api-lite/src/graphql/resolvers/analytics.ts` | Extracted `buildBlocksDashboard(ctx)` from the `getBlocksDashboard` resolver so both the resolver and the new REST route share the same cached node-building logic |
| `packages/api-lite/src/rest/router.ts` | Added `GET /dashboard`, returning `{ nodes }` with `Cache-Control: public, max-age=5` |
| `packages/api-lite/src/server.ts` | Wired `buildBlocksDashboard` into the REST router's `dashboard` dep |
| `docker/vps/nginx.conf.template` | Added `location = /api/dashboard`, copied from `/api/charts` with a 5s `proxy_cache_valid` |
| `packages/app-explorer/src/systems/Home/components/Hero/hooks/useFuelExplorerStatus.tsx` | `useDashboardBlocks` now fetches `/api/dashboard` first, falling back to the GraphQL query on failure |


nelitow:
:+1:

nelitow:
:+1:

PR #719, opened 2026-09-11
fix(api-lite): bound bridge REST log lookups and validate address/from_block

nelitow:
Bounds the three bridge REST lookups (`/bridge/deposit/logs`, `/bridge/block/hashes`, `/bridge/message/relayed/hash`) so a single request can no longer scan and return every matching L1 log row ever indexed.

- `BridgeStore`: every `queryLogs` call now passes `limit: MAX_BRIDGE_ROWS` (1000); `queryLogsForRecipient`'s merged direct+predicate result is re-capped to `MAX_BRIDGE_ROWS` after sorting, since each side being individually capped can still merge to up to 2×`MAX_BRIDGE_ROWS`.
- `router.ts`: `address` on all three bridge routes is now validated with the same `getValidAddress` check the staking routes use (400 on a malformed address); `from_block` on `/bridge/block/hashes` is validated as a non-negative integer (400 otherwise).

Tests: `packages/api-lite/src/bridge/BridgeStore.test.ts`, `packages/api-lite/src/rest/router.test.ts`.


nelitow:
:+1:

nelitow:
:+1:

PR #720, opened 2026-09-11
fix(app-explorer): route the sync status hook through api-lite's /health instead of a direct fuel-core call

nelitow:
## Summary

The sync status indicator used to have every visitor's browser query fuel-core directly every 10 seconds, bypassing api-lite and nginx's rate limit. It now reads the same sync information from api-lite's existing `/health` endpoint, which already tracks fuel-core's tip, and polls every 30 seconds instead of 10. A failing or unhealthy API response now surfaces as an error instead of being read as healthy.

## Changes

| Area | Change |
| --- | --- |
| `packages/app-explorer/src/systems/Core/hooks/useSyncMetrics.ts` | Replaced the direct fuel-core GraphQL call and the indexer GraphQL call with a single fetch to api-lite's `/health`; derives the same sync metrics from `fuelCoreTip`, `servedTip`, and `lag`; raised the poll interval from 10s to 30s and `staleTime` to match; kept the 2s client timeout, now bounding the `/health` fetch; a non-2xx response (including api-lite's 503) now throws instead of being parsed as healthy data. |


nelitow:
:+1:

nelitow:
:+1:

PR #721, opened 2026-09-11
fix(app-explorer): lazy-load bridge and staking pages, split wallet vendor chunk

nelitow:
## Summary

The explorer's production bundle shipped one 7.5 MB JavaScript file to every
visitor, including the wagmi/viem/connectkit/ethers code needed only by the
bridge and staking pages. This change lazy-loads those four pages plus the
transaction overlay dialog, and splits the wallet libraries into their own
cacheable chunk, cutting the main chunk from 2,014 kB to 1,500 kB gzip. The
wallet libraries are still loaded eagerly on every page (via existing global
providers), so the true first-load reduction is smaller — about 2% gzip —
but the new chunk boundary lets browsers cache wallet code separately from
app code that changes more often.

## Changes

| Area | Change |
|---|---|
| `packages/app-explorer/src/App.tsx` | Converted `BridgePage`, `BridgeHistoryPage`, `StakingOnEthereumPage`, `StakingOnFuelPage`, and `OverlayDialog` to `React.lazy`, each wrapped in `Suspense` with a loading fallback |
| `packages/app-explorer/vite.config.ts` | Added `manualChunks` to group `wagmi`, `viem`, `ethers`, `connectkit`, `framer-motion`, `@tanstack/react-query` into a `vendor-wallet` chunk |


nelitow:
:+1:

nelitow:
:+1:

PR #722, opened 2026-09-11
perf(api-lite): serialize each block once and gzip at level 1 in BlockStore

nelitow:
- `BlockStore.load()` now stringifies each freshly loaded block exactly once and reuses that string for both LRU size accounting (`sizeCalculation`/`memorySizes`) and the disk write, instead of stringifying twice per block on the request thread.
- A disk-cache hit with no `normalize` configured reuses the string `readDisk` already decompressed from gzip, instead of re-stringifying the parsed object.
- `writeDisk` now gzips at level 1 (fastest, least compression) instead of the default level 6, since it runs synchronously on the request thread for every new block.
- `HEAP_BYTES_MULTIPLIER` accounting, `sizeOf()`'s raw-byte-length reporting, and eviction behavior are unchanged.
- Tests: `packages/api-lite/src/store/BlockStore.test.ts` — a loader/source block is stringified exactly once per fresh load and zero times on a subsequent disk hit; the on-disk file is a valid level-1 gzip that round-trips to the original block. All 33 pre-existing tests (eviction, `sizeOf()`, normalize, fallback, pinned heights) pass unchanged.


nelitow:
:+1:

nelitow:
:+1:

nelitow:
:+1:

PR #723, opened 2026-09-11
perf(api-lite): decode archive blocks and gzip the disk cache on worker threads

nelitow:
Indexing and serving share one event loop. A backfill batch decodes protobuf, stringifies, and gzips every block on the main thread, and the 2026-08-28 load test measured p97.5 at 8354 ms under backfill against 972 ms from cache.

This adds `DecodeWorkerPool`: `DECODE_WORKERS` worker threads, default one less than the available cores and at least one, that take archive bytes and return the decoded block as JSON text plus the gzipped bytes for the disk cache. Disk-cache hits are gunzipped on a worker too. ArrayBuffers are transferred, not copied. The main thread does one `JSON.parse` per block and no stringify or gzip on the archive path. `DECODE_WORKERS=0` keeps the previous inline path. A worker error or crash rejects that one load with its height, and a crashed worker is replaced; a hung worker is bounded by BlockStore's existing load timeout.

Measured on a synthetic 4.1 MB block, batch of 10 loads: max main-thread event-loop stall 285 to 436 ms inline, 9.9 to 11.3 ms with the pool.

| Change | Where |
|---|---|
| Worker pool, message protocol, transfer helper | `src/store/DecodeWorkerPool.ts` |
| Worker entry, built to `dist/decodeWorker.js` | `src/store/decodeWorker.ts`, `tsup.config.ts` |
| BlockStore uses the pool when given, inline otherwise | `src/store/BlockStore.ts` |
| `DECODE_WORKERS` config and wiring | `src/config.ts`, `src/main.ts` |
| Tests | `DecodeWorkerPool.test.ts`, `BlockStore.test.ts`, `config.test.ts`, `test/helpers/sampleBlock.ts` |

Next step, not in this PR: SQLite writes stay on the main thread. better-sqlite3 connections cannot be shared across threads, and opening a second connection inside a worker changes the locking story between the indexer and the readers, so that move needs its own design.

Deployment: `availableParallelism()` reports host cores, not a container's CPU quota, and each worker is a V8 isolate whose heap sits outside `--max-old-space-size`. `docker/vps/docker-compose.yml` and `docker-compose.prod.yml` therefore set `DECODE_WORKERS=1` for their 0.8-CPU, 384 MB-heap api service; `deploy.md` and `.env.example` document the variable next to the other api settings.

## Fix round 1

Applied on top of 612990f in the worktree, not committed. Changed files: `docker/vps/docker-compose.yml`, `docker/vps/docker-compose.prod.yml`, `docker/vps/.env.example`, `docker/vps/deploy.md`, `packages/api-lite/src/store/BlockStore.test.ts`, `packages/api-lite/src/store/DecodeWorkerPool.test.ts`.

1. `DECODE_WORKERS: "1"` added to the api `environment:` block of both compose files, with a why-comment in `docker-compose.yml` and a pointer to it in `docker-compose.prod.yml`. `.env.example` gets `DECODE_WORKERS=1` with a comment noting the compose files override it, matching the `MEMORY_CACHE_BYTES` note above it. `deploy.md` adds `--set "DECODE_WORKERS=1"` to the api `railway variables` block plus a comment describing the variable and the default's blind spot.
2. `BlockStore.test.ts`: the size-1 pool is built in `beforeAll`, so a `-t` filter that skips the group never spawns a thread.
3. `DecodeWorkerPool.test.ts` "spreads work across workers": after the four `decode()` calls the test asserts `busy.size === 2` and `idle.length === 0`, which holds because dispatch is synchronous, then `idle.length === 2` after the batch resolves. Two distinct workers held tasks at once.

Commands and output:

```
cd <wt>/packages/api-lite && npx jest --config jest.config.ts src/store/DecodeWorkerPool.test.ts src/store/BlockStore.test.ts
Test Suites: 2 passed, 2 total
Tests:       42 passed, 42 total
Time:        1.493 s

cd <wt>/packages/api-lite && npx jest --config jest.config.ts src/store/BlockStore.test.ts -t "fetches once" --detectOpenHandles
Tests:       35 skipped, 1 passed, 36 total      # no open-handle report, process exited

cd <wt>/packages/api-lite && npx biome check src/store/DecodeWorkerPool.test.ts src/store/BlockStore.test.ts
Checked 2 files in 9ms. No fixes applied.

cd <wt>/docker/vps && cp .env.example .env && docker compose -f docker-compose.yml config | grep -n "DECODE_WORKERS\|cpus"; rm .env
10:          cpus: 0.8
21:      DECODE_WORKERS: "1"
57:          cpus: 0.2
REGISTRY_OWNER=x docker compose -f docker-compose.prod.yml config | grep -n "DECODE_WORKERS\|cpus"
7:          cpus: 0.8
18:      DECODE_WORKERS: "1"
48:          cpus: 0.2
```

The temporary `.env` was removed after the render; `git status` shows only the six files above.


nelitow:
:+1:

nelitow:
:+1:

PR #724, opened 2026-09-11
chore(vps): run api-lite as non-root, lock its runtime deps, drop duplicate Google Fonts, add security headers

nelitow:
## Summary

Four independent hardening changes to the api-lite/explorer VPS deployment:

- api-lite's container now starts as root only long enough to `chown` the
  Railway `DATA_DIR` volume mount to the unprivileged `node` user, then
  drops privilege via `su-exec` before running the app - previously it ran
  as root the whole time.
- api-lite's runtime `node_modules` are now installed from a committed
  lockfile (`npm ci`) instead of `npm install` against a floating,
  generated `package.json` - transitive versions no longer drift between
  builds.
- The explorer no longer loads Inter/GeistMono from Google Fonts; it
  self-hosts them (this also required wiring up an existing but previously
  unused self-hosted fonts file - see the report for why).
- nginx now sends `X-Content-Type-Options`, `Referrer-Policy`,
  `X-Frame-Options`, `Strict-Transport-Security`, and a report-only
  Content-Security-Policy. Report-only, not enforcing: `index.html`'s
  inline Vercel Analytics shim script violates `script-src 'self'` as-is,
  so it'll show up in violation reports immediately - that's expected, and
  enforcement needs that inline script removed/nonced first, as a
  follow-up.

**Action required before/at this deploy:** the Railway volume mounted at
`/data` for the `api` service must be writable by uid 1000. The new
entrypoint chowns it automatically on every container start, so a fresh or
already-writable volume needs nothing extra. If the existing production
volume is currently root-owned (likely, since the old image ran as root),
the very first deploy with this image will still work - the entrypoint's
chown runs before the app starts, as root, specifically to fix this - but
if that chown were ever skipped (e.g. by removing the entrypoint later
without re-adding an equivalent step), the app would fail to open its
SQLite file. Confirmed this recovery path in an isolated test (see the
report): a volume `chown`'d to root beforehand was chowned back to
`node:node` by the entrypoint on the next container start.

## Changes

| Area | Change |
| --- | --- |
| `docker/vps/Dockerfile.api-lite` | `deps` stage: `COPY` a new committed lockfile and use `npm ci --omit=dev` instead of `npm install --omit=dev`. `runtime` stage: install `su-exec`, copy in a new entrypoint script, `COPY --chown=node:node` the app's files at build time, and switch from `CMD` alone to `ENTRYPOINT ["/entrypoint.sh"]` + `CMD ["node", "dist/main.js"]`. |
| `docker/vps/entrypoint.sh` (new) | Chowns `$DATA_DIR` (default `/data`) to `node:node` on every container start, then `exec su-exec node "$@"` to drop from root to the `node` user before running the app. |
| `docker/vps/api-lite-runtime.package-lock.json` (new) | Lockfile for api-lite's minimal runtime `package.json`, generated with `npm install --package-lock-only --omit=dev`; regeneration steps documented in a Dockerfile comment. |
| `packages/app-explorer/index.html` | Removed the `fonts.googleapis.com`/`fonts.gstatic.com` preconnect and stylesheet `<link>` tags. |
| `packages/app-explorer/src/index.css` | Added `@import` of `packages/ui/src/theme/fonts.css` - previously unimported anywhere, so removing the Google Fonts links would have left no font-face declaration at all. |
| `packages/ui/src/theme/fonts.css` | Added a second `@font-face` aliasing the Google Fonts name `"Geist Mono"` to the same self-hosted file, since `app-explorer`'s `globals.css` references that name rather than the `"GeistMono"` alias already defined here. Also widened all three `@font-face` blocks from `font-weight: normal` to `font-weight: 100 900` - both files are variable fonts, and a single-value weight descriptor was making the browser synthesize fake bold/light instead of using the real weight axis. |
| `docker/vps/nginx.conf.template` | Added `X-Content-Type-Options`, `Referrer-Policy`, `X-Frame-Options`, `Strict-Transport-Security`, and a report-only `Content-Security-Policy` at server level and repeated them in the four `location` blocks that already set their own `add_header` (`/assets/`, `/api/charts`, `/api/dashboard`, `/`) (nginx doesn't inherit server-level `add_header`s into a location that sets any of its own). `connect-src` extended beyond the given base list with the ETH RPC fallback hosts (Alchemy, Infura, cloudflare-eth.com, llamarpc.com, thirdweb) and Coinbase Wallet SDK hosts found by grepping the app sources. |


nelitow:
:+1:

nelitow:
:+1:

PR #725, opened 2026-09-14
chore(vps): cap the api container at 1 GiB so page cache stops driving the Railway memory bill

nelitow:
## Summary

Railway bills memory on the container's cgroup usage, which includes the kernel page cache for the volume, not only the process. The api's own process uses 270 to 415 MB RSS (from the `/health` lines in the deploy log, 2026-09-14), but the container's memory metric climbed from 0.44 GB to a 14.12 GB peak once the indexer started writing again on 2026-09-11 (about 5.5 GB/day of SQLite rows plus about 13 GB/day of gzipped block cache files). That metric is what the memory line on the usage page charges for: 16.25 USD of the 18.00 USD spent this month so far, 24.56 USD estimated.

This sets a 1 GiB container memory limit in the api's config-as-code file. Under a limit the kernel reclaims the page cache, so the billed figure stays at the limit while the process keeps its 2.4x headroom over the measured peak RSS.

## Changes

| Area | Change |
|------|--------|
| Railway api config | `deploy.limitOverride.containers.memoryBytes` set to 1 GiB (1073741824) |
| Deploy notes | Explains the page-cache accounting, the measured numbers, and that the limit needs the config file path set or the dashboard's Resource Limits, since no service variable carries it |

Deployment: the 1 GiB limit is already applied to the api service through the `serviceInstanceLimitsUpdate` GraphQL mutation (2026-09-14) and `DECODE_WORKERS=1` is set there; this PR records the same limit in `railway.api.json` for a deployment that uses the config file path.


nelitow:
:+1:

PR #726, opened 2026-09-14
fix(api-lite): retry a failed RPC block fetch once and log one line for a failing height

nelitow:
## Summary

A small fraction of block fetches from fuel-core fail because the upstream socket drops mid-request; on testnet, 15 of about 10,800 fetches failed this way over 3 hours, and every one succeeded again on the very next attempt. The RPC block fetch now retries once, after a short pause, only for that kind of transient network failure -- not for an HTTP-level error such as a bad request or a malformed response. Separately, when a block still fails after the retry, the block store's error log is now a single line instead of a full multi-line stack trace, keeping the service log readable; the full stack is still available behind an existing debug flag.

## Changes

| Area | Change |
| --- | --- |
| RPC block fetch (`RpcBlockSource.ts`) | Added an `isTransientNetworkError` check and a single retry, after a 250ms pause and a fresh rate-limit slot, for a dropped-connection-style failure only. |
| Block store logging (`BlockStore.ts`) | `getRange`'s failure log is now one line naming the error type, message, and cause code, with the full stack shown only when the `LOG_S3` debug flag is set. |
| Tests | Added coverage for the retry behavior (success-after-retry, double failure, no retry on an HTTP-shaped error, and that the retry still respects the rate limit) and for the one-line failure log format. |


nelitow:
:+1:

nelitow:
:+1:

PR #728, opened 2026-09-15
fix(api-lite): log a failed CoinGecko price fetch and send an optional demo api key

nelitow:
## Summary

USD amounts (\`amountInUsd\`, \`feeInUsd\`, the fee chart) are null on both Railway apis and the price client hid the reason: every CoinGecko failure was swallowed without a log line. This logs the failure and lets a CoinGecko demo key be sent, since the free endpoint refuses many cloud egress addresses.

| File | Change |
|---|---|
| \`packages/api-lite/src/fuelcore/PriceClient.ts\` | Options object constructor. A non-2xx response, an empty body or a thrown fetch logs one line with the status or error message; a fetch runs at most once per ttl (10 min), so the log line is bounded the same way. When \`apiKey\` is set, the request carries \`x-cg-demo-api-key\`. Last good price still survives a failure. |
| \`packages/api-lite/src/config.ts\` | Optional \`COINGECKO_API_KEY\`. |
| \`packages/api-lite/src/main.ts\` | Passes the key to the client. |
| \`packages/api-lite/src/fuelcore/PriceClient.test.ts\` | Constructor calls updated; new tests for the header and for the 429 log line. |

Tests: 30 passed (PriceClient, config). \`ts:check\`: 0 errors.

Rollout: deploy to testnet, read the log line to learn the real cause, then set \`COINGECKO_API_KEY\` on both api services if it is an address block.


nelitow:
:+1:

nelitow:
:+1:

PR #730, opened 2026-09-15
fix(api-lite): sweep retention in short chunks so the hourly delete no longer stalls the event loop

nelitow:
## Summary

On the mainnet api the hourly retention sweep deleted about a million rows in one synchronous transaction per table. better-sqlite3 runs that on the event loop, so for 70 to 130 s nothing else ran: TipTracker ticks were abandoned, block loads hit the 30 s abandon timeout, the backfill read those as "block missing in S3" and stopped, and the explorer answered 504 (deployment log 2026-09-15, every stall at hh:45 to hh:46 alongside a `retention: N rows` line). Backfill progress fell to 0.19 blocks/s over 10 hours.

| File | Change |
|---|---|
| \`packages/api-lite/src/index/Index.ts\` | \`deleteRange(lo, hi)\` deletes one height slice per transaction and moves \`indexed_from\` past it; \`minHeight()\`; \`freelistPages()\`; \`vacuum(pages)\` frees a bounded number of pages per call. \`deleteBelow\` keeps its signature and delegates. |
| \`packages/api-lite/src/index/Indexer.ts\` | \`retention()\` is async: deletes in 100-block chunks and vacuums 2000 pages at a time, yielding to the event loop between steps; a sweep that is still running skips the next tick. |
| \`packages/api-lite/src/main.ts\` | The hourly interval awaits the sweep and logs a failure. |
| Tests | \`deleteRange\` slice semantics; the size loop now asserts chunk bounds and the final floor. |

Tests: 483 passed. \`ts:check\`: 0 errors.


nelitow:
:+1:

nelitow:
:+1:

PR #731, opened 2026-09-15
fix(app-explorer): keep the home page up when a stats or dashboard response has no data

nelitow:
## Summary

While the mainnet api stalled, the home page crashed into the root error boundary with \`Cannot read properties of undefined (reading 'nodes')\` from \`HomePage\`. The statistics, rolling-stats and dashboard readers dereferenced \`data.statistics.nodes\` and \`getBlocksDashboard.nodes\` without a guard, so one degraded response took the whole page down instead of one tile.

| File | Change |
|---|---|
| \`systems/Home/components/Hero/actions/get-statistics.ts\` | Returns \`null\` when the response has no statistics. |
| \`systems/Home/components/Hero/actions/get-rolling-stats.ts\` | Same. |
| \`systems/Home/components/Hero/hooks/useFuelExplorerStatus.tsx\` | The dashboard readers return \`[]\` when the REST body or the GraphQL fallback has no nodes. |

The hooks already handle null and empty inputs (\`statistics?.\`, \`data?.rollingStats60s ?? {...}\`, \`nodes?.map(...) || []\`), so the tiles render empty and refetch on their own interval.


nelitow:
:+1:

nelitow:
:+1:

PR #732, opened 2026-09-15
chore(api-lite): log slow sqlite calls, event-loop delay, and disk fsync latency

nelitow:
## Why

api-mainnet blocks its event loop for about 60 s several times a day (13:45 and 14:03 UTC today) with zero CPU and no network traffic. The 13:56 UTC retention sweep made it worse, but the 13:45 and 14:03 stalls happened with no sweep running, so the sweep is not the only cause. Nothing in the logs says which synchronous call is blocking.

## What

- `logSlowStatements(db, label)` wraps `prepare`, `exec`, `pragma` and `transaction` on each better-sqlite3 connection (index, cosmos, hot, l1). Any call that takes 1 s or more logs `slow sqlite (<label>) <ms>ms: <sql>`.
- The minute status line gains `loopMaxMs`, the maximum event-loop delay measured by `monitorEventLoopDelay` during that minute.
- A 4 KB write plus fsync on `DATA_DIR/.probe` runs every 15 s off the event loop and logs `disk probe: 4 KB write+fsync took <ms>ms` when it takes 1 s or more.

No behaviour change for requests.

## Test

`pnpm ts:check`, biome, and the index/hot/cosmos/l1 jest suites (89 tests) pass.


nelitow:
:+1:

PR #733, opened 2026-09-15
fix(api-lite): seek cursor counts from the cursor instead of scanning the whole window

nelitow:
## Why

#732 (stall diagnostics) caught the call that freezes api-mainnet:

```
14:34:33 slow sqlite (index) 62550ms: SELECT count(*) AS c FROM (SELECT 1 FROM txs WHERE height >= ? AND height <= ? AND (height > ? OR (height = ? AND tx_index > ?)) LIMIT ?)
```

`newerTxCount` bound the primary-key seek at the retention window start (`height >= from`) and only filtered rows against the cursor, so every page of the transactions list walked all rows older than the cursor. On mainnet that is up to 11M `txs` rows read cold from the volume, about 60 s with zero CPU. `newerCountForAccount` had the same shape for `tx_accounts`.

## What

- `txNewerCount`: the lower bound is now `max(from, cursor.height)`, and the cursor filter is `height > ? OR tx_index > ?`. EXPLAIN QUERY PLAN: `SEARCH txs USING PRIMARY KEY (height>? AND height<?)` starting at the cursor.
- `acctNewerCount`: adds `height >= cursor.height` so the plan becomes `SEARCH tx_accounts USING PRIMARY KEY (account=? AND height>?)`.

Results are unchanged; only the seek start moves.

## Test

Index and transactions resolver suites (38 tests), ts:check, biome.


nelitow:
:+1:

PR #734, opened 2026-09-15
feat(api-lite): store txs and tx_accounts in per-day tables so retention drops tables instead of deleting rows

nelitow:
## Why

The hourly retention sweep deletes rows by height from `tx_accounts`, whose primary key is `(account, height DESC, tx_index DESC)`. A height range maps to random pages across the whole table, and the incremental vacuum after it moves pages one at a time. On mainnet one sweep is about 3600 blocks and 1M `tx_accounts` rows; the 13:57 UTC sweep today blocked the event loop for minutes (tip 502/504, fuel-core and cosmos fetches timing out). #730 (chunked retention) did not help because every chunk still does random page writes.

## What

- `txs` and `tx_accounts` become one table pair per 86 400 heights (about a day on mainnet): `txs_p<N>` and `tx_accounts_p<N>`, created on first write. `blocks` stays one table; its contiguous range deletes stay chunked.
- Retention drops a pair with `DROP TABLE` once every height in it is below the floor. No vacuum: freed pages stay on the freelist and are reused. `fileBytes()` now reports live bytes (page count minus freelist) and the `INDEX_MAX_BYTES` loop drops the oldest pair per pass.
- Reads iterate the live pairs newest first (about `retentionDays + 2` tables): `heightForTx`, `accountExists`, `txsForAccount`, `countForAccount`, `newerCountForAccount`, `txCount`, `newerTxCount`. The account `after` page query now seeks from the cursor height instead of scanning the account's history.
- `deleteAboveRange` only touches pairs that hold a height above `indexed_to`, probed on the `txs_p<N>` primary key first, since the new `tx_accounts_p<N>` tables carry no height index.

## Migration

A database that still has the old `txs`/`tx_accounts` tables serves them as one extra partition covering the min..max height they hold at first open (`meta.legacy_tx_range`). Heights inside that range keep writing to the old tables so no height lands in two partitions. The pair is dropped like any other once its max height is below the floor. Nothing is rewritten at boot.

Design notes: see the sections above. The spec file is kept out of this PR because policy-bot needs a second approver for `docs/` paths.

## Rollout

Deploy api (testnet) first, verify, then api-mainnet. After the legacy pair drops on mainnet, put api-mainnet back to `INDEX_RETENTION_DAYS=2` and `INDEX_MAX_BYTES=12000000000`.

## Test

Full api-lite suite: 486 tests pass, including new tests for cross-partition history/cursors/counts, partition drop, and the legacy migration on a file-backed database. `ts:check` and biome pass.


nelitow:
:+1:

nelitow:
:+1:

nelitow:
:+1:

nelitow:
:+1:

PR #735, opened 2026-09-15
fix(api-lite): read the legacy tx range with one aggregate per statement

nelitow:
## Why

The first boot after #734 (per-day tx tables) on testnet logged:

```
slow sqlite (index) 23552ms: SELECT MIN(height) AS lo, MAX(height) AS hi FROM txs
```

SQLite only takes the primary-key min/max shortcut when a query has a single aggregate; with two it scans the table. The mainnet `txs` table is several times larger, so its first boot would spend minutes there.

## What

`loadPartitions` reads MIN and MAX in two statements. Same result, read from the two ends of the primary key.

## Test

Index suites pass (48 tests), ts:check and biome pass.


nelitow:
:+1:

PR #736, opened 2026-09-15
fix(api-lite): pin the tx partition bucket size in the database

nelitow:
## Why

#734 (per-day tx tables) derives a partition's height range from its table name: `txs_p<N>` covers `N * bucketBlocks` to `(N+1) * bucketBlocks - 1`. Changing `DEFAULT_BUCKET_BLOCKS` in a later release would reinterpret every existing table as a range it was never written with. Reads filter those partitions out by range, writes create a second table for the same heights, and the rows go silently missing with no error. Every test builds a fresh index, so none of them can catch it.

## What

The bucket size is written to `meta.bucket_blocks` when the first bucket partition is created, and a database that has it uses the stored value, logging one line when it differs from the configured one.

## Test

Index suites pass (49 tests), including a new case that reopens a file-backed database with a different bucket size and checks the existing partition keeps its range and still takes new writes. `ts:check` and biome pass.


nelitow:
:+1:

PR #737, opened 2026-09-15
ci: give the explorer production build a 4 GB heap in every e2e job

nelitow:
## Why

`E2E Tests Soft` and `E2E Tests api-lite` have failed on every pull request today (#731 through #736), both in the same step:

```
Build production version of explorer
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
Failed: app-explorer#build:prod
```

Both jobs run on `warp-ubuntu-latest-x64-4x`, where node picks a 2 GB default heap; the crash is at about 2005 MB. The sharded `E2E Tests` job runs the identical build and passes because its `8x` runner defaults to 4 GB. Nothing about the failing jobs' own work is at fault, and neither is a required check, so the failures have been sitting red and ignored.

## What

All three build steps set `NODE_OPTIONS: --max-old-space-size=4096`, so the two 4x jobs get the same budget the 8x job already had and every job is on one explicit number instead of a per-runner default.

## Test

The two jobs run on this pull request.


nelitow:
:+1:

PR #738, opened 2026-09-15
chore(explorer): drop Vercel Web Analytics, the script only exists on Vercel

nelitow:
## Summary

The explorer renders Vercel's Web Analytics component, which loads a script that only Vercel's edge serves. On the Railway deployment the request falls through to the SPA fallback, so every page logs "Failed to load script from /_vercel/insights/script.js" in the console and no analytics are recorded. Analytics for the Railway deployment come from Cloudflare instead, so the component has nothing left to do.

Removing it clears the console error and drops one dependency. Nothing else in the app uses the package. Typecheck reports the same 254 pre-existing errors as main, none in the changed file.

## Changes

| Area | Change |
|------|--------|
| App entry point | Removed the Vercel analytics component and its import |
| Dependencies | Removed the @vercel/analytics package and refreshed the lockfile |


nelitow:
👍

nelitow:
👍

nelitow:
👍

PR #739, opened 2026-09-16
fix(vps): serve routes that share a name with a build directory

nelitow:
## Summary

The explorer's nginx config falls back to the app for unknown paths, but it first tried the path as a directory. The build output contains `assets` and `fonts` directories, so requests for `/assets` and `/fonts` answered with a redirect into those directories instead of the app, and the redirect pointed at an `http://` address because Railway terminates TLS in front of nginx. app-testnet.fuel.network shows this today.

Dropping the directory attempt makes those paths fall through to the app, which is what Vercel serves. Real files are unaffected: hashed assets and the two self-hosted font files still return 200 from their own locations.

Checked against a local build in the nginx image: /assets, /fonts, /tx, /account/<id>/assets and unknown paths all return the app with 200, /assets/<hashed file> and /fonts/InterVariable.woff2 return 200.

## Changes

| Area | Change |
|------|--------|
| nginx config | Stopped resolving request paths as directories, so a route named like a build directory reaches the app |
| nginx config | Turned off absolute redirects, which would otherwise point at http:// behind Railway's TLS |


nelitow:
👍

PR #743, opened 2026-09-17
fix(api-lite): number account history pages from one fuel-core list

nelitow:
## Summary

Account transaction pages showed page numbers that jumped around: the foundation wallet went from "61 - 70 of 70" to "1 - 10 of 70" to "1 - 10 of 1000+", although it has 456 transactions. Every page now reads its numbers from one list of the account's transaction ids from fuel-core, so the label counts down by 10 per page against one fixed total. This also fixes the "newer" arrow on busy accounts, which returned the previous page shifted by one row.

## Changes

| Area | Change |
|------|--------|
| Page numbers | Index pages and fuel-core pages take their start, end and total from the account's newest 1001 transaction ids on fuel-core, cached like the other fuel-core pages |
| Busy accounts | Accounts whose index already holds 1001 or more rows keep their index numbering and skip the fuel-core list |
| Newer arrow | An `after` page now returns the rows next to the cursor instead of dropping the nearest one |
| fuel-core client | New `txIdsByOwner` query that selects ids and cursors only, since fuel-core rejects a large page that also selects status as too complex |
| Sparse rows below the index window | Account paging stops at the oldest contiguous index row and continues from fuel-core, because blocks decoded for an older page write rows for every account they contain |
| Unknown totals | A page whose rows the fuel-core list cannot confirm reports the cap, which the explorer renders as "1000+", instead of a wrong exact number |
| Failures | If the fuel-core list cannot be fetched, the page keeps its previous numbers |

Validated headed on a local stack reading mainnet (R2 blocks, mainnet fuel-core) with Playwright: the foundation wallet walks all 46 pages from "447 - 456 of 456" down to "1 - 6 of 456" and back, and a busy account walks 5 pages and back with matching rows and numbers.


nelitow:
:+1:

PR #745, opened 2026-09-21
fix(staking): unstick withdrawals left on sequencer sync and the dead Finalize button

nelitow:
## Summary

Two staking withdrawal bugs reported by support. Withdrawals sent after 2026-09-17 21:12 UTC never left "Synchronizing with Sequencer", because the sequencer poller in api-lite stopped on a request that never finished and stayed stopped without logging anything. Separately, the Finalize button on a ready withdrawal did nothing, because a failed price lookup kept the status dialog in its loading state forever.

## Changes

| Area | Change |
|------|--------|
| Withdraw status dialog | A failed `/assets_rate` request no longer blocks the dialog. Finalize works again; only the USD label is missing when prices fail. The other 9 staking dialogs already behaved this way. |
| Sequencer poller (api-lite) | Each sequencer request now has a hard 20 s deadline on top of the abort signal, so one hung request ends that poll cycle and is logged instead of stopping the poller for good. |
| `/health` (api-lite) | Adds `cosmos.tipAt`, the time the sequencer tip was last read. During the outage `/health` showed `cursor == tip` and `ok: true` for 4 days; a stale `tipAt` makes that state visible. |
| Tests | Two poller tests cover a tip response and a block response that never arrive. Both hung before the fix. |

The stalled pollers resume when the `api-mainnet` and `api` (testnet) services restart, which a deploy of this change does. A plain restart before the merge has the same effect. Mainnet then has about 57,000 sequencer blocks to catch up.

Closes O2T-1866


PraneshASP (approved):
lgtm, thanks!

PR #746, opened 2026-09-21
fix(app-staking): let the withdraw status dialog reach idle when the event fetch fails

nelitow:
## Summary

Follow-up to #745. The withdraw status dialog reloads the withdrawal every 60 seconds. If one of those reloads failed, the dialog stayed in its loading state and ignored the Finalize click, the same symptom #745 fixed for the price lookup. A failed reload now records the error and the dialog keeps working, then retries on the next 60 second reload.

## Changes

| Area | Change |
|------|--------|
| Withdraw status dialog | A failed event fetch now ends in a `failure` state and stores the error, matching the stake status dialog. The dialog's existing error selectors read that value; before this nothing ever set it. |

Part of O2T-1866


nelitow:
👍

PR #747, opened 2026-09-21
fix(app-portal): load bridge history when the indexer API base is a relative path

nelitow:
## Summary

The bridge history page on app.fuel.network showed "No activity yet" with a "Failed to construct 'URL': Invalid URL" toast, so users could not find or execute their withdrawals to Ethereum. Production sets the indexer API base to the relative path `/api`, and the bridge built its request addresses in a way that only accepts a full address. The bridge now resolves a relative base against the page address, so history and withdrawal execution load again.

## Changes

| Area | Change |
|------|--------|
| Bridge deposit history | Request address accepts a relative indexer API base |
| Bridge withdrawal execution | Relayed message lookup accepts a relative indexer API base |
| Bridge block commit lookup | Committed block hashes lookup accepts a relative indexer API base |
| Shared URL helpers | New `absoluteUrl` helper that resolves a relative base against the page origin |


nelitow:
:+1:

PR #750, opened 2026-09-23
fix(api-lite): find accounts whose history is older than the index window

nelitow:
## Summary

Searching for a valid account address returned "No results found" when the account had no activity inside api-lite's rolling index window. The account page itself worked, because it reads balances from fuel-core. Search now falls back to fuel-core's owner history, so any account with at least one transaction is found. A random address still returns nothing.

## Changes

| Area | Change |
|------|--------|
| Search resolver | When the local index finds no block, contract, transaction, account or predicate, ask fuel-core for the owner's last transaction and return the account if one exists |
| Tests | Covers an account known only to fuel-core, and an unknown address that must still resolve to null |


nelitow:
👍

